PR #6, opened 2024-05-09
updated Warg and refactored

calvinrp:


crates/wasm-pkg-loader/src/source/warg.rs
```diff
@@ -88,22 +94,26 @@ impl PackageSource for WargSource {
...
-            .boxed())
+        let package_name = package.try_into()?;
+
+        // warg client already validated the digest
+        let download = self
+            .client
+            .download_exact(&package_name, &release.version)
+            .await?;
+        let file = tokio::fs::File::open(download.path).await?;
+        Ok(ReaderStream::new(file).map_err(Into::into).boxed())
```
lann, line 116: Is this change necessary? It isn't always desirable to write to disk...

crates/wasm-pkg-loader/src/config.rs
```diff
@@ -44,6 +44,10 @@ impl ClientConfig {
         self
     }
 
+    pub fn get_default_registry(&self) -> Option<&str> {
```
lann, line 47: `get_` is discouraged: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

crates/wasm-pkg-loader/src/main.rs
```diff
@@ -20,11 +22,30 @@ async fn main() -> anyhow::Result<()> {
...
+                _ => {}
+            };
         }
+
+        // if not configured and not using Warg home URL as default registry,
+        // then set the `wasi` default registry
+        if is_missing_config && config.get_default_registry().is_none() {
+            config.namespace_registry("wasi", "bytecodealliance.org");
+        };
+
```
lann, line 48: Let's discuss this in https://github.com/bytecodealliance/wasm-pkg-tools/discussions/3

crates/wasm-pkg-loader/src/source/warg.rs
```diff
@@ -61,7 +62,13 @@ impl PackageSource for WargSource {
...
             .releases()
-            .map(|release| release.version.clone())
+            .filter_map(|release| {
+                if release.yanked() {
+                    None
+                } else {
+                    Some(release.version.clone())
+                }
+            })
             .collect())
```
lann, line 72: I wonder if this should be configurable, e.g. an `include_yanked` options.

crates/wasm-pkg-loader/src/config.rs
calvinrp: Yeah, there was a name conflict. I'll refactor the other names to have `set_` prefix.

crates/wasm-pkg-loader/src/source/warg.rs
calvinrp: Or maybe label yanked in the output list?

lann: Yeah but it also might make sense to allow (attempting to) fetching the actual artifact too.

crates/wasm-pkg-loader/src/source/warg.rs
calvinrp: I'll refactor to ignore the Warg cache

calvinrp: Is it a problem that Warg caches the package logs? Or do you need entirely in-memory?

calvinrp: The solution feels like adding an `InMemoryClient` instead of using `FileSystemClient` in `warg_client`. Thoughts?

lann: There is no (durable) writable disk available in this scenario; caching is handled at a higher level. You're probably right that a separate `InMemoryClient` would make sense as well, though its a little less important for the relatively small registry metadata.

calvinrp: Should this crate CLI never use file system Warg download cache? Or should it be configuration?

crates/wasm-pkg-loader/src/main.rs
calvinrp: Just caught up on that thread. I think this is in inline with what was discussed there.

If global `wasm-pkg-tools` config file is not defined but the global `warg` config file is defined, then it falls back to using the `warg` config given this tool can use the `warg_client`.

If there is no global config for either, it uses `bytecodealliance.org`.

crates/wasm-pkg-loader/src/source/warg.rs
lann: It just needs some path that doesn't require the entire package content to be written to disk.

crates/wasm-pkg-loader/src/config.rs
calvinrp: Refactored, let me know if that is OK now.

crates/wasm-pkg-loader/src/main.rs
lann: Having two different sources for "default registry" interacting this way seems potentially confusing. It might make sense to consolidate config between `warg` and this library a bit to end up with a similar UX; that's what I'm trying to hash out in the discussion thread.

crates/wasm-pkg-loader/src/source/warg.rs
calvinrp: Pushed an update will list all versions and mark yanked ones.

crates/wasm-pkg-loader/src/source/warg.rs
calvinrp: Fixed with the updated Warg 0.6.0

crates/wasm-pkg-loader/src/main.rs
calvinrp: Complete agreement. Do you think this implementation makes sense for this PR? And then we work on the rest in follow on PRs?

lann: I don't think we want to land this kind of fundamental config logic without discussion: https://github.com/bytecodealliance/wasm-pkg-tools/discussions/10

calvinrp: @lann I think we have a path forward on the config, from our discussions. As for this PR and this config, how should we proceed?

lann: Can we just revert this chunk of the PR?

calvinrp: @lann Reverted

PR #16, opened 2024-05-15
Add wkg crate

### LINKED ISSUE #5: Add `wasm-pkg` CLI tool
This would be modeled after `wasm-tools`. One initial subcommand could be `wasm-pkg loader`, taking over for the separate `wasm-pkg-loader` bin.

lann:
Closes #5


lann:
Reserved https://crates.io/crates/wkg

PR #20, opened 2024-05-17
oci: Make all docker credential errors non-fatal

### LINKED ISSUE #19: `wasm-pkg-loader` oci backend fails if no docker config
The error handling logic for `docker_credentials` doesn't properly handle missing or malformed `~/.docker/config.json`, causing even requests that don't need auth to fail with `Error: failed to get registry credentials: Unable to read config`

Workaround: `mkdir -p ~/.docker && echo '{}' > ~/.docker/config.json`

lann:
The docker_credential crate doesn't provide a ton of precision in its errors, so we'll downgrade all of its errors to a warning at worst to prevent unnecessary failures. This should remove an entire class of spurious failures at the expense of slightly harder debugging in certain cases.

Fixes #19

### DIFF
--- a/crates/wasm-pkg-loader/src/source/oci.rs
+++ b/crates/wasm-pkg-loader/src/source/oci.rs
@@ -128,14 +128,18 @@ impl OciSource {
                     "identity tokens not supported"
                 )));
             }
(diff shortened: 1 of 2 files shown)


lann:
cc @FrankReh

calvinrp (approved):

PR #21, opened 2024-05-17
Prepare 0.3.0

lann:


.gitignore
```diff
@@ -1,4 +1,3 @@
 target/
-Cargo.lock
```
lann, line 2: This seems to be standard practice for workspaces with both library and binary crates.

fibonacci1729 (approved):

PR #23, opened 2024-05-22
Support `protocol = "http"` in OCI registry config

itowlson:
This is useful/required for local registries e.g.

```toml
default_registry = "localhost:5000"

[registry."localhost:5000"]
type = "oci"
protocol = "http"
```

### DIFF
--- a/crates/wasm-pkg-loader/src/config.rs
+++ b/crates/wasm-pkg-loader/src/config.rs
@@ -73,13 +73,10 @@ impl ClientConfig {
     pub fn set_oci_registry_config(
         &mut self,
         registry: impl Into<String>,
-        client_config: Option<OciClientConfig>,
+        client_config: OciClientConfig,
         credentials: Option<BasicCredentials>,
     ) -> Result<&mut Self, Error> {
-        if client_config
-            .as_ref()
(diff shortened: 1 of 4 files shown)


crates/wasm-pkg-loader/src/config.rs
```diff
@@ -73,13 +73,10 @@ impl ClientConfig {
...
-        client_config: Option<OciClientConfig>,
+        client_config: OciClientConfig,
         credentials: Option<BasicCredentials>,
     ) -> Result<&mut Self, Error> {
-        if client_config
-            .as_ref()
-            .is_some_and(|cfg| cfg.platform_resolver.is_some())
-        {
+        if client_config.platform_resolver.is_some() {
             Error::InvalidConfig(anyhow::anyhow!(
```
itowlson, line 80: I was a bit puzzled by this line. It looks like it constructs the Error and then drops it rather than returning it. But when I tried changing it to return the Error, the CI failed with this error message. So I've left it along but I feel like I'm missing something...!

lann: That is definitely broken!

PR #25, opened 2024-05-28
Add wasm-pkg-common crate

lann:


calvinrp:
Just getting around to reviewing, but still going thru.

I think we can safely drop the backward compatible well-known logic / fields. We aren't relying on the behavior in our install instructions yet.

lann:
> I think we can safely drop the backward compatible well-known logic / fields.

We can drop it in a future change; it just reduces the amount of coordination needed between various moving pieces.

lann:
I was going to wait until I had some integration work done with this, but given that #29 depends on this now I'll just rebase what exists now for review.

thomastaylor312 (commented):
Everything looks good to me

thomastaylor312 (approved):

calvinrp (approved):

PR #26, opened 2024-05-31
ref(wasm-pkg-loader): turn default features off for tracing-subscriber crate

vdice:
This PR disables the default features for the tracing-subscriber crate, primarily to avoid bringing in the [tracing-log](https://crates.io/crates/tracing-log) crate, which can introduce issues in downstream consumers because it attempts to init a global logger which may already be initialized (ref https://github.com/spinkube/containerd-shim-spin/issues/61).

If this project does have a need to utilize the tracing-log crate, we could also go an opt-out approach.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1781,16 +1781,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "nu-ansi-term"
-version = "0.46.0"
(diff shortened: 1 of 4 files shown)


lann (approved):
lgtm; follow up: #27

PR #28, opened 2024-05-31
feat(*)!: Updates pkg-loader to use oci-wasm crate

thomastaylor312:
This adds support for the new oci-wasm crate and uses the types and client wrapper from that crate. This is marked as a breaking change because it will only support the new layer content type of `application/wasm` rather than the older format. I figured that would be fine since this is a new crate.


thomastaylor312:
If we'd like to make it backwards compatible to the old layer and config type, we could probably add that in the crate as an option, but I figure it might be better to rip the bandaid

thomastaylor312:
I'll figure out the test thing soon!

crates/wasm-pkg-loader/.vscode/settings.json
thomastaylor312, line 1: Note that I removed this directory because it was IDE specific and added it to gitignore

lann (approved):
lgtm. I'd just like to republish the bytecodealliance.org packages with the correct config before merging.

lann:
> If we'd like to make it backwards compatible to the old layer and config type, we could probably add that in the crate as an option, but I figure it might be better to rip the bandaid

Nah. That was all just hedging my bets before the cncf thing had solidified.

calvinrp (approved):

thomastaylor312:
Sounds great @lann. Is there a problem with merging and then not cutting a release until you republish those packages?

lann:
Just that I might forget :sweat_smile:

lann:
I'll either fix the packages or set reminder(s) for myself and merge anyway by eod tomorrow.

lann:
Oh it looks like the oci-wasm crate/repo is still just owned by you @thomastaylor312; I think we should get those under some shared ownership before adding it as a dependency here to make sure we can land bug fixes etc.

lann:
Also @thomastaylor312 I'm looking at updating those packages; how have you been testing this? Just `oras`?

thomastaylor312:
@lann I already added @jsturtevant as another maintainer on the project and [I finished all the steps](https://github.com/bytecodealliance/governance/issues/88#issuecomment-2142620863) after approval to move to BCA. So I think we're ok there at this point

As for pushing, I've been using the latest main of `wash` to push things since it can push any component. You can also use the code I did in the e2e tests in this PR if you want to whip up something quick. Otherwise, I am working on adding push in #31

lann:
> @lann I already added @jsturtevant as another maintainer on the project

Cool; could you add him as a crate owner too?

thomastaylor312:
Yeah I can. I was mostly waiting so I could make a GH team have permissions, but I can do that for now. I sent him the invite and he should be accepting soon

jsturtevant:
I've been added as a crate owner.  Thanks!

PR #29, opened 2024-06-03
ref(*): Switches all tracing macros to use fields

thomastaylor312:
This is a small update to make sure we're following best practice and logging the values as fields

### DIFF
--- a/crates/wkg/src/main.rs
+++ b/crates/wkg/src/main.rs
@@ -86,7 +86,7 @@ impl GetCommand {
             None => {
                 println!("No version specified; fetching version list...");
                 let versions = client.list_all_versions(&package).await?;
-                tracing::trace!(?versions);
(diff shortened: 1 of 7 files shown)


thomastaylor312:
Please note that I didn't change the fields in one of the OCI functions since I already did that in #28. Should save on rebasing time

crates/wkg/src/main.rs
```diff
@@ -200,7 +200,6 @@ async fn main() -> anyhow::Result<()> {
         .init();
 
     let cli = Cli::parse();
-    tracing::debug!(?cli);
```
thomastaylor312, line 203: I removed this because we're likely to have credentials or other sensitive things in this CLI so just removing it for now. We can add back in some more specific debug logging as this evolves

calvinrp (approved):

PR #30, opened 2024-06-04
ref(wkg): Cleanup of various parts of the wkg code

thomastaylor312:
This cleans up some usage of blocking IO in an async context as well as a few other small things

### DIFF
--- a/crates/wkg/src/main.rs
+++ b/crates/wkg/src/main.rs
@@ -1,12 +1,12 @@
 mod package_spec;
 
-use std::{io::Seek, path::PathBuf};
+use std::path::PathBuf;
 
 use anyhow::{ensure, Context};
 use clap::{Args, Parser, Subcommand, ValueEnum};
 use futures_util::TryStreamExt;
 use package_spec::PackageSpec;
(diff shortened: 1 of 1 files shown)


calvinrp (approved):

lann:
This looks fine but I'm not sure what makes these changes necessary?

thomastaylor312:
Blocking IO in an async context can cause deadlocks and other weird bugs (ask me how I know!). Also, I think generally going back and cleaning up some code is always a good reason for a PR 😄

lann:
Ah ok. I just ask because I've been freely mixing async and sync here under the assumption that it shouldn't make any difference for this sort of cli tool 🤷

thomastaylor312:
It _shouldn't_ but I've been bit one too many times

thomastaylor312:
I'd really like to follow best practices here if we can unless we're sure we're ok with taking that risk on

thomastaylor312:
@lann I did some more looking and I retract my previous statement. I think because these are running serially, we're ok. We might need to use them in the other crates since they're libraries and could be used in parallel with other IO bound tasks

PR #31, opened 2024-06-04
feat(*)!: Adds wasm-pkg-pusher to repository

thomastaylor312:
This PR adds a common pushing library to wasm-pkg-tools and also adds push functionality into `wkg`. This involved refactoring out common OCI code into the common crate


thomastaylor312:
Heads up that I created this as the eventual PR, but I have not yet added pushing (I'll do that in a commit tomorrow). Mostly I created this as a heads up for @lann since I am building on top of his new common crate. I personally think there is enough of the common stuff in there that we could just merge #25 as is and then this PR before continuing on with the refactor.

Also, note that this is based on top of #28 as well so you can pretty much skip all of those commits to see what I am doing. This was rebased so that the order is

#25 -> #28 -> this PR

lann:
Ok I'll try to get #25 merged today or tomorrow

thomastaylor312:
Going to close this in favor of a new PR I am opening soon

Issue #13, opened 2024-05-15
Unified publishing


lann:
Currently, if you want to publish a package to a warg registry you use `warg` and if you want to publish to an OCI registry you use some OCI tool (like `oras`). For toolchain-specific tools like `cargo-component` that want to support their own `publish` flow it would be nice to have a unified publishing library available.

Supporting the full feature set of e.g. warg seems out of scope for this library, but it does seem plausible to support a subset of common operations like basic single release publishing, selecting the appropriate protocol by consulting the well-known wasm-pkg URL for the destination registry domain.

This functionality could also be exposed via an e.g. `wkg publish` subcommand.

thomastaylor312:
This should be completed in #80

PR #33, opened 2024-06-08
feat(*)!: Add support for pushing and pulling OCI artifacts

thomastaylor312:
This PR adds push and pull support for wasm artifacts to wkg. Essentially `wkg get` and `wkg push` are set up to be CLI wrappers around the various registry clients. This is purposeful because pushing an artiface and pulling it (for inspection/composition/etc) is a fundamentally different task than "loading" a package, which is done for dependency fetching. One of them is done via fully qualified URLs and possibly multiple layers of nesting depending on security setups and team structure. Loading a dependency on the other hand expects a flat-ish namespace of components to pull for satisfying dependencies (a la crates.io or pkg.go.dev). With that in mind, there are 3 key use cases people need for developer workflows that we should support here:

- Pushing built wit packages or components to a registry. We need a standard set of tools that can do this in the BCA, and `wkg` is that place
- Pulling components for debugging, inspection, republishing, and more. Similarly to pushing, we need this in the standard tool chain
- Loading packages as a dependency from a registry. This is so things like `cargo component` can fetch something by package reference (i.e. `wasi:http@0.2.0`)


crates/wkg/src/warg.rs
```diff
@@ -0,0 +1,23 @@
+use clap::Args;
+
+#[derive(Args, Debug)]
+pub struct GetArgs {
+    // TODO: Args
```
thomastaylor312, line 5: @calvinrp I could use your help implementing the warg part of this since you'll probably be quicker than me for defining the right set of flags and what not. If you can't get to it, I'll get to it next week

thomastaylor312:
Please note that this stacked on top of #34, so filter the commit list down to see what I am changing here. It is in draft because I haven't finished implementing the warg part of the new `wkg` commands

thomastaylor312:
One more note from over the weekend thinking: I think there should also be a `wkg publish` which is pairs with `wkg load`. That will use all the config and mapping to push an artifact to a registry being used as a dependency registry

lann:
> here should also be a `wkg publish`

https://github.com/bytecodealliance/wasm-pkg-tools/issues/13 :slightly_smiling_face:

thomastaylor312:
Closing in favor of #44

PR #34, opened 2024-06-08
ref(*)!: Refactors loader to use the new common package

thomastaylor312:
This removes a bunch of no longer needed code and removes the binary for wasm-pkg-loader in favor of wkg.


crates/wasm-pkg-common/src/registry.rs
```diff
@@ -151,6 +168,32 @@ impl RegistryMetadata {
                 .map_err(Error::InvalidRegistryMetadata)?,
         ))
     }
+
+    /// Creates a new [`RegistryMetadata`] by attempting to fetch it from a well known path. Only
+    /// errors if an error occurred with the request itself. If the URL was not found, `None` is
+    /// returned
+    pub async fn fetch(domain: &str) -> Result<Option<Self>, Error> {
```
lann, line 175: I'm not opposed to having this as a convenience but `reqwest` is a relatively heavyweight dependency so I'd like to gate it behind a non-default feature e.g. `metadata-client` or something.

crates/wasm-pkg-common/Cargo.toml
```diff
@@ -8,10 +8,17 @@ authors.workspace = true
...
+semver = { workspace = true }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+tokio = { workspace = true, features = ["rt", "sync"] }
 toml = "0.8.13"
 thiserror = "1.0"
-tracing = "0.1"
\ No newline at end of file
+tracing = "0.1"
+warg-client = { workspace = true }
```
lann, line 24: It looks like some (?) of these new dependencies aren't actually used in this crate.

crates/wasm-pkg-loader/src/source/mod.rs
```diff
@@ -6,7 +6,6 @@ use std::cmp::Ordering;
 
 use crate::{Error, PackageRef, Release};
 
-pub mod local;
```
lann, line 9: Why is this removed?

lann (commented):
Partial review

crates/wasm-pkg-common/Cargo.toml
thomastaylor312: Oh yep. I had pulled something in and then backed that out and forgot to remove deps

crates/wasm-pkg-common/src/config/mod.rs
```diff
@@ -142,8 +142,8 @@ impl Config {
     }
 
     /// Sets the default registry.
-    pub fn set_default_registry(&mut self, registry: Option<Registry>) {
-        self.default_registry = registry;
+    pub fn set_default_registry(&mut self, registry: Registry) {
+        self.default_registry = Some(registry);
```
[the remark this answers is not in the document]
thomastaylor312: Oh, I misread it then. I'll revert that

crates/wasm-pkg-common/src/registry.rs
```diff
@@ -64,6 +65,20 @@ pub const REGISTRY_METADATA_PATH: &str = "/.well-known/wasm-pkg/registry.json";
...
+}
+
+/// The configuration returned by an OCI registry
+#[derive(Deserialize)]
+pub struct OciProtocolConfig {
+    pub registry: String,
+    #[serde(rename = "namespacePrefix")]
+    pub namespace_prefix: Option<String>,
+}
+
```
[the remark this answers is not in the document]
thomastaylor312: So were you thinking this would just be a private struct in the loader crate for now? Want to make sure I put this in the right place

crates/wasm-pkg-common/src/registry.rs
thomastaylor312: Yeah, that makes sense to me. I wondered about some of the heavy deps as well

crates/wasm-pkg-loader/src/source/mod.rs
thomastaylor312: I hadn't seen any examples or anything around that showed this and the loader code so I had assumed this was just an artifact for testing. Can add back

thomastaylor312: Should the type name for this just be "local" or was there something else you had in mind?

thomastaylor312:
Ok, I've done most of the changes. Tomorrow I'll work on adding local back and moving the configs to the right spot

crates/wasm-pkg-common/src/registry.rs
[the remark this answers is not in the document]
lann: Private struct in the loader crate probably makes sense for now. We'll probably dedupe the warg config into `warg-client` and I imagine we'll end up with a `wasm-pkg-oci` crate later.

crates/wasm-pkg-loader/src/source/mod.rs
lann: Yeah this was requested in a registry meeting to make sure we can handle different kinds of sources. I think "local" is fine in this context.

lann:
I pulled a large subset of the changes from this PR into #41 to prevent conflicts with other work I'm doing.

thomastaylor312:
Closing this because most was merged in #41

PR #44, opened 2024-06-17
feat(wkg)!: Adds support for pushing and pulling OCI wasm artifacts

thomastaylor312:
This adds support for pushing and pulling compatible OCI wasm artifacts from the command line. As part of this refactor, it makes the original loading feature a separate command for use in debugging. Some of the features from the `load` subcommand should probably make their way into the push and pull command, but for now just keeping it simple

### DIFF
--- a/crates/wkg/src/main.rs
+++ b/crates/wkg/src/main.rs
@@ -9,6 +9,10 @@ use wasm_pkg_client::Client;
 use wasm_pkg_common::{config::Config, package::PackageSpec, registry::Registry};
 use wit_component::DecodedWasm;
 
+mod oci;
+
+use oci::OciCommands;
+
 #[derive(Parser, Debug)]
 #[command(version)]
 struct Cli {
@@ -19,18 +23,22 @@ struct Cli {
 #[derive(Args, Debug)]
 struct RegistryArgs {
     /// The registry domain to use. Overrides configuration file(s).
-    #[arg(long = "registry", value_name = "DOMAIN")]
+    #[arg(long = "registry", value_name = "REGISTRY", env = "WKG_REGISTRY")]
     registry: Option<Registry>,
 }
 
 #[derive(Subcommand, Debug)]
+#[allow(clippy::large_enum_variant)]
 enum Commands {
-    /// Get a package.
-    Get(GetCommand),
+    /// Load a package. This is for use in debugging dependency fetching. For pulling a component, use `wit get`
+    Get(GetArgs),
+    /// Commands for interacting with OCI registries
+    #[clap(subcommand)]
+    Oci(OciCommands),
 }
 
 #[derive(Args, Debug)]
-struct GetCommand {
+struct GetArgs {
     /// Output path. If this ends with a '/', a filename based on the package
     /// name, version, and format will be appended, e.g.
     /// `name-space_name@1.0.0.wasm``.
@@ -61,7 +69,7 @@ enum Format {
     Wit,
 }
 
-impl GetCommand {
+impl GetArgs {
     pub async fn run(self) -> anyhow::Result<()> {
         let PackageSpec { package, version } = self.package_spec;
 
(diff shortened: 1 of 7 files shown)


crates/wkg/src/oci.rs
```diff
@@ -0,0 +1,185 @@
...
+        value_delimiter = ','
+    )]
+    pub insecure: Vec<String>,
+}
+
+/// Commands for interacting with OCI registries
+#[derive(Debug, Subcommand)]
+pub enum OciCommands {
+    /// Get a component from an OCI registry and write it to a file.
+    Get(GetArgs),
```
yoshuawuyts, line 74: You referred to this as "pull" in the PR description. I liked that because it also matches the terminology used by the OCI distribution crate, the Docker CLI, and Oras. Can we call this command "pull"?

thomastaylor312: Good catch. I had kept previous terminology from #33 where I was doing something for both warg and OCI

lann: I intentionally avoided calling this `pull`. While `wkg oci pull` is clearly similar to `oras pull`, `wkg get` is intended to do more than just pull a single artifact; it is more analogous to something like `go get` or `npm install`.

crates/wkg/src/main.rs
```diff
@@ -19,18 +23,22 @@ struct Cli {
 #[derive(Args, Debug)]
 struct RegistryArgs {
     /// The registry domain to use. Overrides configuration file(s).
-    #[arg(long = "registry", value_name = "DOMAIN")]
+    #[arg(long = "registry", value_name = "WKG_DOMAIN")]
```
lann, line 26: It looks like you were maybe thinking the `value_name` here was an `env`?

crates/wkg/src/oci.rs
thomastaylor312: So these commands are all under `wkg oci`. Does that change anything? From what @calvinrp conveyed to me, we wanted to have those under a separate oci subcommand and then we have `load` (can be renamed to `get`) and `publish` at the top level. Did I misunderstand the intent?

crates/wkg/src/main.rs
thomastaylor312: Whoops that was definitely me rebasing poorly

crates/wkg/src/oci.rs
lann: Ohh sorry I looked at the diff for this change and assumed it was the existing command. No, `pull` is good here.

crates/wkg/src/main.rs
```diff
@@ -19,18 +23,22 @@ struct Cli {
...
     registry: Option<Registry>,
 }
 
 #[derive(Subcommand, Debug)]
+#[allow(clippy::large_enum_variant)]
 enum Commands {
-    /// Get a package.
-    Get(GetCommand),
+    /// Load a package. This is for use in debugging dependency fetching. For pulling a component, use `wit get`
+    Load(LoadArgs),
```
lann, line 34: I don't feel too strongly about `get` vs `load`, though `get` is at least familiar to go devs. :shrug:

calvinrp: Leaning towards keeping it as `wkg get` since it is short and easier to remember (at least for me).

thomastaylor312: Ok, I'll swap that out

thomastaylor312: Changed it to get!

crates/wkg/src/main.rs
```diff
@@ -105,7 +113,7 @@ impl GetCommand {
         };
 
         let (tmp_file, tmp_path) =
-            tempfile::NamedTempFile::with_prefix_in(".wkg-get", parent_dir)?.into_parts();
+            tempfile::NamedTempFile::with_prefix_in(".wkg-load", parent_dir)?.into_parts();
```
calvinrp, line 116: Change this one back as well?

yoshuawuyts:
Oh by the way, I tried this out today with a colleague and it seems to work really well.

It's really nice to be able to push and pull `.wasm` files directly to an OCI registry. I don't have any concrete nits or anything on this PR, but figured it would still be useful to share that: "I tried this and it worked well!"

PR #53, opened 2024-06-21
Code reorganization and doc improvements

lann:
Slicing this out of some ongoing work since it seems independently useful/reviewable:

- Move ContentDigest to wasm-pkg-common
- Add a bunch of doc comments, including a loader example
- Reorganize loader "sources"
  - expose backend registry config structs


lann:
In light of #54 I'm going to close this in favor of a PR implementing that rename.

PR #58, opened 2024-06-25
feat(caching): Adds a new caching client

### LINKED ISSUE #45: Add caching for OCI loading
`oci-distribution` doesn't include any caching of data by default. We should add some caching that stores the pulled artifacts in `$XDG_CACHE_DIR` somewhere

thomastaylor312:
This adds a trait so it can be extendible and provides a default filesystem implementation. I chose this way of doing it because I didn't think the hassle of having a `Client<T>` was worth it just to add caching. This is set up in such a way that you can still use a caching client as a normal client

Closes #45


crates/wasm-pkg-client/src/caching/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+    cache: T,
+}
+
+impl<T> AsRef<Client> for CachingClient<T> {
+    fn as_ref(&self) -> &Client {
+        &self.client
+    }
+}
+
+impl<T> Deref for CachingClient<T> {
```
lann, line 40: Using Deref this way is generally discouraged: https://rust-unofficial.github.io/patterns/anti_patterns/deref.html

thomastaylor312: I can go either way on this. From that link:

> There is no one good alternative.

Which makes me laugh because if it is an antipattern, then hopefully they'd have something else you can do 😆 

The purpose here was not to have polymorphism so much as a thin wrapper around the normal client to avoid the need of doing something like `client.client.method` since you'll still need to do things like fetch the `Release`. 

I am not married to this though. If you're strongly opposed it is easy enough to drop. Just let me know!

lann: For me, the important part is "this is a surprising idiom"; I was indeed confused at first about how this was supposed to work. `Deref` is _mostly_ used only for "smart pointer"-style generic wrappers; things that add actual functionality more often expose inner data with `AsRef` or full delegation.

thomastaylor312: Ok, I removed it

calvinrp (approved):

PR #62, opened 2024-06-28
fix: Resolve docker hub credentials

### LINKED ISSUE #61: Credentials to docker hub are not resolved automatically
When executing `oci push myaccount/image:latest file.wasm`, I have encountered an auth error:
```
Error: Unable to push image

Caused by:
    Server error: url https://index.docker.io/v2/myaccount/image/blobs/uploads/, code: 401, message: {"errors":[{"code":"UNAUTHORIZED","message":"authentication required","detail":[{"Type":"repository","Class":"","Name":"myaccount/image","Action":"pull"},{"Type":"repository","Class":"","Name":"myaccount/image","Action":"push"}]}]}
```
It does work when passing `-u` and `-p`. I've debugged the issue to the fact that the auth is stored under `"https://index.docker.io/v1/"` but  `https://docker.io` is passed to `docker_credential::get_credential`, preparing a PR.

tomasol:
Fix resolving credentials from docker `config.json`. Add a new `server` parameter variant for
`docker_credential::get_credential`.

When using the default registry, it must be set to `https://index.docker.io/v1/`.
Retain the generic lookup for other registries.

Fixes #61

### DIFF
--- a/crates/wkg/src/oci.rs
+++ b/crates/wkg/src/oci.rs
@@ -36,16 +36,20 @@ impl Auth {
         match (self.username, self.password) {
             (Some(username), Some(password)) => Ok(RegistryAuth::Basic(username, password)),
             (None, None) => {
-                let server_url = format!("https://{}", reference.registry());
-                match docker_credential::get_credential(&server_url) {
-                    Ok(DockerCredential::UsernamePassword(username, password)) => {
-                        return Ok(RegistryAuth::Basic(username, password));
-                    }
-                    Ok(DockerCredential::IdentityToken(_)) => {
-                        return Err(anyhow::anyhow!("identity tokens not supported"));
-                    }
-                    Err(err) => {
-                        tracing::debug!("Failed to look up OCI credentials: {err}");
+                let server_url = format!("https://{}", reference.resolve_registry());
+                for server_url in [&server_url, &format!("{server_url}/v1/")] {
+                    match docker_credential::get_credential(server_url) {
+                        Ok(DockerCredential::UsernamePassword(username, password)) => {
+                            return Ok(RegistryAuth::Basic(username, password));
+                        }
+                        Ok(DockerCredential::IdentityToken(_)) => {
+                            return Err(anyhow::anyhow!("identity tokens not supported"));
+                        }
+                        Err(err) => {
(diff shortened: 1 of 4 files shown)


crates/wkg/src/oci.rs
```diff
@@ -36,16 +36,20 @@ impl Auth {
...
-                    Ok(DockerCredential::UsernamePassword(username, password)) => {
-                        return Ok(RegistryAuth::Basic(username, password));
-                    }
-                    Ok(DockerCredential::IdentityToken(_)) => {
-                        return Err(anyhow::anyhow!("identity tokens not supported"));
-                    }
-                    Err(err) => {
-                        tracing::debug!("Failed to look up OCI credentials: {err}");
+                let server_url = format!("https://{}", reference.resolve_registry());
+                for server_url in [&server_url, &format!("{server_url}/v1/")] {
```
thomastaylor312, line 40: Is docker the only one that forces the `v1` ending? If so, it might be clearer to just check if it is docker and then append. A comment here would also clarify. Took me a second when I first looked

thomastaylor312 (commented):
Thanks for the fix! Just one small question around making the code a little more clear.

crates/wkg/src/oci.rs
tomasol: I've looked into the sources of `docker login` and did not find any other cases where `/v1/` would be appended, also according to [this](https://jfrog.com/help/r/jfrog-artifactory-documentation/test-your-docker-and-artifactory-cloud-setup):
>You cannot use a context path when providing the registry path (e.g localhost:8082/artifactory is not valid)

So I've replaced the for loop with a 1:1 mapping as you suggested.

tomasol:
I was testing the PR and realized that even the non-default registries are keyed incorrectly, fixed in the last commit. For reference, here is my `~/.docker/config`
```json
{
        "auths": {
                "ghcr.io": {
                        "auth": "redacted"
                },
                "https://index.docker.io/v1/": {
                        "auth": "redacted"
                },
                "registry.fly.io": {
                        "auth": "redacted"
                }
        }
}

```

thomastaylor312 (approved):
Thanks for digging into this thoroughly! LGTM

PR #66, opened 2024-07-08
fix(client): Implement various helpers for serialzing config structs

thomastaylor312:
Without these, it was basically impossible to serialize back to the intended config format expected

### DIFF
--- a/crates/wasm-pkg-client/src/oci/config.rs
+++ b/crates/wasm-pkg-client/src/oci/config.rs
@@ -5,7 +5,7 @@ use base64::{
 };
 use oci_distribution::client::ClientConfig;
 use secrecy::{ExposeSecret, SecretString};
-use serde::Deserialize;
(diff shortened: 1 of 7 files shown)


crates/wasm-pkg-client/src/oci/config.rs
```diff
@@ -32,6 +32,25 @@ impl Clone for OciRegistryConfig {
     }
 }
 
+impl Serialize for OciRegistryConfig {
```
lann, line 35: This is helpful when you're just converting to another type and immediately serializing: https://serde.rs/container-attrs.html#into

lann (approved):

crates/wasm-pkg-client/src/oci/config.rs
thomastaylor312: Oh that's cool. I've never used that. Let me give it a whirl

PR #72, opened 2024-07-12
fix(warg): Removes call to cache for versions

thomastaylor312:
The implementation for listing versions was pulling from the cache. This caused it to miss updates to versions when being used in cargo-component.

I discovered this while working on cargo-component and it was blocking tests from working

### DIFF
--- a/crates/wasm-pkg-client/src/warg/mod.rs
+++ b/crates/wasm-pkg-client/src/warg/mod.rs
@@ -54,6 +54,16 @@ impl WargBackend {
         package: &PackageRef,
     ) -> Result<PackageInfo, Error> {
         let package_name = package_ref_to_name(package)?;
+        // NOTE(thomastaylor312): We need to make sure we're up to date with all packages, but if we
+        // bypass the cache every time, we'll have to fetch the whole package log every time rather
+        // than loading from cache on disk. The remaining question here is the performance impact.
+        // At scale, we don't know if this will result in a lot of HTTP requests even though the
+        // packages were updated on a previous call. This should be good enough for now, but we
+        // might need to revisit this later.
+        self.client
+            .update()
+            .await
+            .map_err(|e| Error::RegistryError(e.into()))?;
         self.client
             .package(&package_name)
             .await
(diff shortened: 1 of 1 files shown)


thomastaylor312:
Note that I also pushed the fix for non-https urls here from #68 so it was decoupled from that

crates/wasm-pkg-client/src/warg/mod.rs
```diff
@@ -35,13 +35,23 @@ impl WargBackend {
...
         let WargRegistryConfig {
             client_config,
             auth_token,
             ..
         } = registry_config.try_into()?;
 
+        let url = warg_meta.url.unwrap_or_else(|| {
+            // If we just pass registry plain, warg will assume it is https. This is a workaround to
+            // assume that a local registry is http.
+            if registry.host() == "localhost" || registry.host() == "127.0.0.1" {
```
calvinrp, line 48: add `"0.0.0.0"`?

calvinrp:
Could you point me to where in `cargo-component` that you are running into the cache issue? I just want to confirm that we should always skip the cache vs explicitly doing a `update()` before `package()` (https://github.com/bytecodealliance/registry/blob/main/crates/client/src/lib.rs#L621-L622).

crates/wasm-pkg-client/src/warg/mod.rs
thomastaylor312: 0.0.0.0 is a weird one because it means all interfaces for some things and I think it is rejected (maybe?) on Windows

thomastaylor312:
Calling `update` is probably the better choice here. Still working my way around the API

thomastaylor312:
@calvinrp So it looks like the update method updates _all_ packages rather than just a specific one being fetched. That feels like it could be possibly expensive operation to run every time someone wants to fetch a package. Is that still the best option? I feel like fetching only for a specific package every time is a bit more predictable. I just don't know if it is going to cause heavy load on a server to list all packages to a client and then update the checkpoints (even if those checkpoints are empty data updates)

calvinrp:
@thomastaylor312 Sorry, I'm just trying to understand why the caching is problem. On how this being called.

thomastaylor312:
@calvinrp and I synced offline. For now we are going to call `update` every time, but I left a comment explaining possible tradeoffs for later

calvinrp (approved):

PR #73, opened 2024-07-15
Add support for annotations to `wkg oci push`

yoshuawuyts:
This adds support for annotations to `wkg oci push`. Annotations can be added via a new `--annotation key=value` flag which can be repeated multiple times to add multiple annotations. For example this can be used to add a standard OCI description to the OCI image of the `wasi:filesystem` spec:

```bash
wkg oci push \
    ghcr.io/yoshuawuyts/wasi/filesystem:0.2.0 \
    ./webassembly_wasi_filesystem.wasm \
    --annotation org.opencontainers.image.description="the WASI filesystem spec"
```

Thanks!

### DIFF
--- a/crates/wkg/src/oci.rs
+++ b/crates/wkg/src/oci.rs
@@ -9,6 +9,7 @@ use oci_distribution::{
     Reference,
 };
 use oci_wasm::{WasmClient, WasmConfig};
+use std::error::Error;
 
 #[derive(Debug, Args)]
 pub struct Auth {
@@ -129,6 +130,24 @@ pub struct PushArgs {
 
     /// The path to the file to push
     pub file: PathBuf,
+
+    /// Add an OCI annotation to the image manifest
+    #[clap(long = "annotation", value_parser = parse_key_val::<String, String>)]
+    pub annotation: Vec<(String, String)>,
+}
+
+/// Parse a single key-value pair
+fn parse_key_val<T, U>(s: &str) -> Result<(T, U), Box<dyn Error + Send + Sync + 'static>>
+where
+    T: std::str::FromStr,
+    T::Err: Error + Send + Sync + 'static,
+    U: std::str::FromStr,
+    U::Err: Error + Send + Sync + 'static,
+{
+    let pos = s
+        .find('=')
+        .ok_or_else(|| format!("invalid KEY=value: no `=` found in `{s}`"))?;
+    Ok((s[..pos].parse()?, s[pos + 1..].parse()?))
 }
 
 impl PushArgs {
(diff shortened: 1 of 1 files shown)


crates/wkg/src/oci.rs
```diff
@@ -129,6 +130,24 @@ pub struct PushArgs {
...
+
+/// Parse a single key-value pair
+fn parse_key_val<T, U>(s: &str) -> Result<(T, U), Box<dyn Error + Send + Sync + 'static>>
+where
+    T: std::str::FromStr,
+    T::Err: Error + Send + Sync + 'static,
+    U: std::str::FromStr,
+    U::Err: Error + Send + Sync + 'static,
+{
+    let pos = s
```
thomastaylor312, line 147: Nit: I think it would be more readable to use `s.split_once` and match on the option instead. I know you can technically split the string and take ownership, but this is a CLI, so to_owned on them should be just fine

crates/wkg/src/oci.rs
```diff
@@ -129,6 +130,24 @@ pub struct PushArgs {
...
     /// The path to the file to push
     pub file: PathBuf,
+
+    /// Add an OCI annotation to the image manifest
+    #[clap(long = "annotation", value_parser = parse_key_val::<String, String>)]
+    pub annotation: Vec<(String, String)>,
+}
+
+/// Parse a single key-value pair
+fn parse_key_val<T, U>(s: &str) -> Result<(T, U), Box<dyn Error + Send + Sync + 'static>>
```
thomastaylor312, line 140: Any reason this needs to be generic? We'll probably only use this to get two strings out, so it could be simpler just to omit the generics and the list of bounds. It also means we don't have to parse and instead can just to_owned the split string.

thomastaylor312: Nit: Also, I am pretty sure this can just be `anyhow::Error` since we use anyhow most places here.

thomastaylor312 (changes requested):
Thanks for adding this! One minor question/request and then this should be good to go

crates/wkg/src/oci.rs
```diff
@@ -9,6 +9,7 @@ use oci_distribution::{
     Reference,
 };
 use oci_wasm::{WasmClient, WasmConfig};
+use std::error::Error;
```
thomastaylor312, line 12: Nit: Can you please put this up with the other `std` imports up top?

crates/wkg/src/oci.rs
yoshuawuyts: Heh, sure I can change it to not be generic. This should honestly probably be a `clap` built-in, since I copied this specific function from their examples since it's apparently a common case.

yoshuawuyts:
tidied it up; I think this should be good now!

thomastaylor312 (approved):

PR #79, opened 2024-08-30
chore(docs): Adds documentation for configuration and repo

thomastaylor312:
This should give people enough information to actually configure things properly on our first release. I also added a `--config` flag to `wkg` so that users can also manually specify a config file

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,185 @@
-<h1 align="center">wasm-pkg-tools</h1>
-<div align="center">
+# wasm-pkg-tools
 
+<div align="center">
   <strong>A <a href="https://bytecodealliance.org/">Bytecode Alliance</a> project</strong>
-  
-  <p>
-    <strong>
-      Tools to package up <a href="https://github.com/webassembly/component-model">Wasm Components</a>
-    </strong>
-  </p>
 </div>
 
+Tools to package up [Wasm Components](https://github.com/webassembly/component-model)
+
+This repo contains several Rust crates that can be used for fetching and publishing Wasm Components
+to OCI or Warg registries. It is also the home of the `wkg` command line tool, which exposes all the
+functionality of the libraries. The first (but not only) focus of this project is allow for fetching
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,18 +1,151 @@
...
+For Warg registries, the JSON looks like this:
+
+```json
+{
+  "wargUrl": "https://warg.wa.dev"
+}
+```
+
+The `wargUrl` field is the base URL of the Warg registry used when connecting the client.
+Namespacing for warg is built in to the protocol.
```
calvinrp, line 132: I think is the old format with the "backward compatible support": https://github.com/bytecodealliance/wasm-pkg-tools/blob/main/crates/wasm-pkg-common/src/metadata.rs#L15-L32

https://github.com/bytecodealliance/wasm-pkg-tools/blob/main/crates/wasm-pkg-common/src/metadata.rs#L90-L98

```json
{
  "preferredProtocol":"warg",
  "warg": {"url":"https://warg.example.com"},
  "oci": {"registry": "oci.example.com", "namespacePrefix": "prefix/"}
}
```

thomastaylor312: This has been updated

README.md
```diff
@@ -1,18 +1,171 @@
...
+of Wit Interfaces stored as components for use in creating components. It can also be used to fetch
+and publish component "libraries" to/from a registry.
+
+## Installation
+
+Right now installation of `wkg` is manual. You can either clone the repo and build from source, or
+download a pre-built binary from the [releases
+page](https://github.com/bytecodealliance/wasm-pkg-tools/releases) and add it to your `PATH`. In the
+future we will add this tool to various package managers. If this is something you'd like to help
+with, please feel free to open some PRs!
```
calvinrp, line 21: `cargo install wkg` should work for the released versions, also...

If you have the [cargo-binstall](https://github.com/cargo-bins/cargo-binstall)
utility installed, `wkg` can also be installed via a prebuilt
release artifact, saving time on the installation:

```
cargo binstall wkg
```

calvinrp (approved):

PR #80, opened 2024-08-31
feat(client): Adds publishing to the client

thomastaylor312:
This PR adds publishing support to the client as well as refactoring the e2e tests to run with `cargo test`.


calvinrp (approved):
I can do follow up PR with a e2e test for Warg

PR #82, opened 2024-09-03
feat(warg): Adds config option for signing secret

thomastaylor312:
Adds and documents a new option for adding a signing key for publishing. This should enable additional flexibility in setting a signing key for tools like `cargo component`

### DIFF
--- a/crates/wasm-pkg-client/src/warg/config.rs
+++ b/crates/wasm-pkg-client/src/warg/config.rs
@@ -1,21 +1,37 @@
-use std::path::PathBuf;
+use std::{fmt::Debug, path::PathBuf, sync::Arc};
 
 use secrecy::{ExposeSecret, SecretString};
 use serde::{Deserialize, Serialize, Serializer};
+use warg_crypto::signing::PrivateKey;
 use wasm_pkg_common::{config::RegistryConfig, Error};
 
 /// Registry configuration for Warg backends.
 ///
 /// See: [`RegistryConfig::backend_config`]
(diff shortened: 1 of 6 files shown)


crates/wasm-pkg-client/src/warg/config.rs
```diff
@@ -1,29 +1,46 @@
...
+    pub signing_key: Option<Arc<PrivateKey>>,
     /// The path to the Warg config file, if specified.
     pub config_file: Option<PathBuf>,
 }
 
+impl Debug for WargRegistryConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("WargRegistryConfig")
+            .field("client_config", &self.client_config)
+            .field("auth_token", &self.auth_token)
```
calvinrp, line 30: Should `"[redacted]"` for the `auth_token` as well?

calvinrp (commented):
Otherwise, looks good!

crates/wasm-pkg-client/src/warg/config.rs
thomastaylor312: I'm letting the `Secret` type handle that for me. I would have still derived `Debug` because `PrivateKey` is just a `Secret` under the hood, but it didn't implement that for me. I can add that if you'd like, but this was mostly a workaround because `PrivateKey` isn't `Debug`

calvinrp (approved):

calvinrp (approved):

Issue #89, opened 2024-09-06
Tell user where to get 'wit'


pothos:
The help output for `wkg get` says to use `wit get` for puling a component. This sentence reads like it is a CLI tool (and not talking about the WIT format) and it would be good to have a link on where to get it. If it's a CLI tool, what repo is `wit` developed in?

calvinrp:
That's a typo in the docs. Should read `wkg get` not `wit get`. Sorry for the confusion.

There is a [`wit` CLI tool ](https://crates.io/crates/wit) for authoring wit packages and publishing. It lives in the [cargo component repo](https://github.com/bytecodealliance/cargo-component/tree/main/crates/wit). But it is slated to be deprecated soon. To be replaced with a `wkg wit` subcommand.

@thomastaylor312 is the doc generated from [this comment](https://github.com/bytecodealliance/wasm-pkg-tools/blob/main/crates/wkg/src/main.rs#L41)?

calvinrp:
Fixed in #90

calvinrp:
Thanks for reporting this, BTW. Definitely something that needed fixing

PR #90, opened 2024-09-06
updated wkg CLI help docs

calvinrp:
Addresses #89


lann (approved):
Should we update `crates/wkg/Cargo.toml`'s [`readme`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-readme-field) with this patch release too?

PR #91, opened 2024-09-25
feat(wkg-core): Adds new wkg-core crate with lockfile support

thomastaylor312:
This adds a new `wit` subcommand to `wkg` that supports building wit packages and fetching/populating a deps directory. I'm sure there is much more we can do here and some obtuse edge cases that aren't supported, but
I did test fetching dependencies for various worlds that used wasi:http and wasi:cli so the basics are functional! In a follow up PR, I'll add some more integration tests


thomastaylor312:
Heads up that the test failure is happening because macos runners don't have docker installed by default. I'll figure that out shortly

crates/wkg-core/src/lock.rs
```diff
@@ -0,0 +1,870 @@
...
+
+impl PartialEq for LockFile {
+    fn eq(&self, other: &Self) -> bool {
+        self.packages == other.packages && self.version == other.version
+    }
+}
+
+impl Eq for LockFile {}
+
+impl Serialize for LockFile {
```
lann, line 53: You can `#[derive(Serialize)]` and then `#[serde(skip)]` individual fields.

thomastaylor312:
Heads up that I added a commit with integration tests. Update isn't tested yet because we need to be pointing at wasi.dev so we can try an update from 0.2.1, but also that doesn't do much yet (it is essentially the same as fetch but forces a new lock file (for example if you changed registries)

calvinrp (approved):

PR #92, opened 2024-09-25
chore(common): Switches default registry to wasi.dev for wasi namespace

thomastaylor312:
I also updated one of the integration tests to use the 0.2.1 interfaces
so we could make sure that works as well


thomastaylor312:
This is stacked on top of #91 so I'll mark as ready once that merges

calvinrp (approved):

PR #113, opened 2024-10-11
fix: Removes use of dirs in favor of etcetera

### LINKED ISSUE #98: Add a `FileCache::global_path` method
I've noticed as I've been integrating in wkg to various tools that I never added something that gets a file cache at the default path. We should probably add a method to do so.

It might also be good to move the `Common` args struct from the `wkg` CLI to `wasm-pkg-core` since those args will pretty much be used anywhere someone is integrating into a CLI (cargo-component, wash, spin, etc)

### LINKED ISSUE #111: [BUG] Config path does not seem to respect `$XDG_CONFIG_HOME` variable
From my (very sparse) testing, it doesn't look like the `wkg` config path is respecting the $XDG_CONFIG_HOME variable.

```console
$ echo "$XDG_CONFIG_HOME"
/Users/lheywood/.config

$ EDITOR=ls wkg config --edit
/Users/lheywood/Library/Application Support/wasm-pkg/config.toml
```console

thomastaylor312:
The `dirs` crate doesn't actually respect XDG vars, so this swaps out to etcetera. This also has the side effect of removing a transitive dep that had an MPL licenses (which can cause headaches for CNCF and other projects).

Also adds a global path for the file cache

Fixes #111
Fixes #98


calvinrp (approved):

thomastaylor312:
Heads up that this will break anyone that has a config setup already, but since that is probably no more than 3-4 people in the world right now, I figured that is fine. Rather do it now before we start merging in the dep to everything

PR #116, opened 2024-10-18
add `--check` mode to `wkg get`

pchickey:
which compares the retrieved contents to an existing file, and exits with 0 if no difference or 1 if there is a difference.

I intend to use this in CI workflows much like https://github.com/bytecodealliance/wasmtime/blob/main/.github/workflows/main.yml#L508-L510

### DIFF
--- a/crates/wkg/src/main.rs
+++ b/crates/wkg/src/main.rs
@@ -199,6 +199,13 @@ struct GetArgs {
     #[arg(long, value_enum, default_value = "auto")]
     format: Format,
 
+    /// Check that the got package matches the existing file at the output
+    /// path. Output path will not be modified. Program exits with codes
+    /// simmilar to diff(1): exits with 1 if there were differences, and 0
+    /// means no differences.
+    #[arg(long)]
+    check: bool,
+
     /// Overwrite any existing output file.
     #[arg(long)]
     overwrite: bool,
@@ -269,6 +276,11 @@ enum Format {
 
 impl GetArgs {
     pub async fn run(self) -> anyhow::Result<()> {
+        ensure!(
+            !(self.overwrite && self.check),
+            "Not allowed to specify both --check and --overwrite"
+        );
+
         let PackageSpec { package, version } = self.package_spec;
         let mut config = self.common.load_config().await?;
         if let Some(registry) = self.registry_args.registry.clone() {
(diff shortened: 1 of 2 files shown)


crates/wkg/src/main.rs
```diff
@@ -199,6 +199,13 @@ struct GetArgs {
     #[arg(long, value_enum, default_value = "auto")]
     format: Format,
 
+    /// Check that the got package matches the existing file at the output
```
thomastaylor312, line 202: Spelling/grammar (normally these are nits, but since it is help text, these should be fixed): "fetched" or "retrieved" instead of "got" and "simmilar" -> "similar"

crates/wkg/src/main.rs
```diff
@@ -269,6 +276,11 @@ enum Format {
 
 impl GetArgs {
     pub async fn run(self) -> anyhow::Result<()> {
+        ensure!(
+            !(self.overwrite && self.check),
```
thomastaylor312, line 280: Can you add [`conflicts_with`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with) to the flags so this can be caught by the CLI?

crates/wkg/src/main.rs
```diff
@@ -371,21 +383,34 @@ impl GetArgs {
...
-        if let Some(wit) = wit {
-            std::fs::write(&output_path, wit)
-                .with_context(|| format!("Failed to write WIT to {output_path:?}"))?
+        if self.check {
+            let existing = std::fs::read(&output_path)
+                .with_context(|| format!("Failed to read {output_path:?}"))?;
+            let latest = if let Some(wit) = wit {
+                wit.into_bytes()
+            } else {
+                std::fs::read(&tmp_path).with_context(|| format!("Failed to read {tmp_path:?}"))?
```
thomastaylor312, line 393: Nit: Should probably use `tokio` methods instead here since we're in async land. This shouldn't cause an issue with the CLI though, hence a nit

thomastaylor312 (changes requested):
This looks great! A few minor fixes needed, but otherwise good to go

crates/wkg/src/main.rs
pchickey: Thanks, I didn't remember that was possible with clap.

crates/wkg/src/main.rs
pchickey: Fixed

crates/wkg/src/main.rs
pchickey: Fixed

lann (approved):

PR #124, opened 2024-10-31
enable cargo-component lock files

macovedj:
These changes bring logic from cargo-component so that dependency resolution can be fully sourced from wasm-pkg-tools in cargo-component. Namely, cargo-component can use wkg.lock instead of Cargo-component.lock


macovedj:
@thomastaylor312 @calvinrp @lann ready for review

vados-cosmonic:
Hey as we've certainly come a long way since this PR (and cargo-component is no longer the suggested way of building Rust components, etc), I'm going to close this out! Happy to re-open if there's some interest in rewriting.

Issue #132, opened 2024-11-21
[BUG] Fetching order causes inconsistent results


brooksmtownsend:
I notice some inconsistent behavior with registry overrides and local dependency overrides when reaching a certain level of dependencies with `wkg wit fetch`. I was able to get a pretty minimal reproduction in https://github.com/bytecodealliance/wasm-pkg-tools/compare/main...brooksmtownsend:wasm-pkg-tools:example/wit-fetch-mismatch?expand=1

worlds for reference:
```javascript
package exampleone:foo@0.1.0;

interface some-foo-thing {
  record request {
    foo: list<u8>,
  }
}

world handler {
  include wasmcloud:component-go/imports@0.1.0;
  import exampletwo:bar/some-bar-thing@0.1.0;

  export some-foo-thing;
}
package examplethree:baz@0.1.0;

interface some-baz-thing {
  record requestresponse {
    baz: list<u8>,
  }
}

world handler {
  include wasmcloud:component-go/imports@0.1.0;

  export some-baz-thing;
}
package exampletwo:bar@0.1.0;

interface some-bar-thing {
  record response {
    bar: list<u8>,
  }
}

world handler {
  include wasmcloud:component-go/imports@0.1.0;
  import examplethree:baz/some-baz-thing@0.1.0;

  export some-bar-thing;
}
```

This is a fairly simple example where a `wit-one` world depends on `wit-two` interface which depends on the `wit-three` interface. I'm able to fetch the dependencies for wit-three (just a wasmCloud upstream interface) just fine, same with `wit-two`, but then trying to fetch the interface for `wit-one` will fail _sometimes_. It's unclear to me if all of the worlds having the same interface as a dependency also affects this. 

To reproduce, you can check out the example in my fork and `cd examples/wit-deps-conflict-wit-one` and run:
```bash
wkg wit fetch --config ../package_config.toml
```

Depending on the run ~50% of the time, it completes successfully. Other times, I get an error that there's no registry configured for "examplethree":
```bash
➜ wkg wit fetch --config ../package_config.toml

➜ wkg wit fetch --config ../package_config.toml
Error: no registry configured for namespace "examplethree"
```

[...]

PR #135, opened 2024-12-05
fix(core): allow transitive fetching local deps

brooksmtownsend:
- **fix(core): use local override over registry dep**
- **test(core): add test for transitive local dep**
- **deps: enable tracing-subscriber ansi feature**

Fixes https://github.com/bytecodealliance/wasm-pkg-tools/issues/132

This PR:
1. Fixes an issue with transitive dependency fetching where a local dependency of a local dependency was treated as a registry dependency, which would fail
1. Adds a test for the above case to ensure that this works
1. Enables the `ansi` feature of tracing-subscriber so we can all have colored logging output <3

### DIFF
--- a/crates/wasm-pkg-core/src/resolver.rs
+++ b/crates/wasm-pkg-core/src/resolver.rs
@@ -407,6 +407,8 @@ impl<'a> DependencyResolver<'a> {
                     _ => None,
                 };
 
+                // So if it wasn't already fetched first? then we'll try and resolve it later, and the override
+                // is not present there for some reason
                 if !force_override
                     && (self.resolutions.contains_key(name) || self.dependencies.contains_key(name))
                 {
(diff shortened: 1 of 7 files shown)


crates/wasm-pkg-core/src/resolver.rs
```diff
@@ -423,17 +425,29 @@ impl<'a> DependencyResolver<'a> {
...
+                // This is a bit of a hack, but if there are multiple local dependencies that are
+                // nested and overriden, getting the packages from the local package treats _all_
+                // deps as registry deps. So if we're handling a local path and the dependencies
+                // have a registry package already, override it. Otherwise follow normal overrides.
+                // We should definitely fix this and change where we resolve these things
+                let should_insert = force_override
+                    || self.dependencies.contains_key(name)
+                    || !self.resolutions.contains_key(name);
+                if !should_insert {
+                    tracing::debug!(%name, "2dependency already exists and registry override is not set, ignoring");
```
joonas, line 442: Looks like you might've had an accidental `2` sneak into the debug message here

Cargo.toml
```diff
@@ -33,6 +33,7 @@ tokio-util = "0.7.10"
 toml = "0.8"
 tracing = "0.1.40"
 tracing-subscriber = { version = "0.3.18", default-features = false, features = [
+    "ansi",
```
joonas, line 36: Just my $0.02, it feels like this should be set in [`wkg` Cargo.toml](https://github.com/bytecodealliance/wasm-pkg-tools/blob/main/crates/wkg/Cargo.toml#L26) and not at the workspace level.

joonas (approved):

thomastaylor312 (approved):

Issue #118, opened 2024-10-25
`wkg get -o some_dir/` gives confusing error if `some_dir` doesn't exist


itowlson:
```
$ wkg get fermyon:spin@3.0.0 -o WIT-TEST2/ --format wit --registry registrytest-vfztdiyy.fermyon.app
Getting fermyon:spin@3.0.0...
Error: No such file or directory (os error 2) at path "/home/ivan/testing/s3regtest/WIT-TEST2/.wkg-getT7LRZK"
```

The actual problem is that `WIT-TEST2` doesn't exist (and wkg is reluctant to create it), but the temporary file name sends me in the wrong direction.  It would be more helpful if `wkg` either:

* created the directory; or
* printed e.g. "directory WIT-TEST2 does not exist"

PR #141, opened 2024-12-29
Create the output directory if it does not exist (#118)

chikoski:
This PR is a fix for issue #118. With this fix, the `get` command creates the directory specified by the `-o` option if the specified directory does not exist. 

For example, the following command creates the `wit` directory and saves the `wasi:http@0.2.0.wit` in it:

```zsh
$ wkg get -o wit/ wasi:http0.2.0
```


crates/wkg/src/main.rs
```diff
@@ -1,4 +1,4 @@
-use std::{io::Seek, path::PathBuf};
+use std::{fs, io::Seek, path::PathBuf};
```
thomastaylor312, line 1: Can you please swap out to use `tokio::fs` since the function is an async one?

thomastaylor312 (dismissed):
Thanks for adding this! Just one small change needed

crates/wkg/src/main.rs
chikoski: Apologies for my delayed response. Updated the code as suggest.

Note: the CI is failed as wasm32-wasi is not available now.

PR #142, opened 2025-01-12
Update target for CI with wasm32-wasip1 as wasm32-wasi is removed

chikoski:
This PR updates the target that CI uses with `wasm32-wasip1`, as `wasm32-wasi` is removed.

Note: This PR is a part of #141, as the CI failure blocks it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
         with:
-          targets: "wasm32-wasi"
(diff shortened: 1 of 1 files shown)


thomastaylor312:
Thanks for fixing this @chikoski! So the problem you're running into here is that the test fixtures have cargo configs that set the target (like [here](https://github.com/bytecodealliance/wasm-pkg-tools/blob/55c0f404dda4313c7f8a7c42c951c91a28d67d40/crates/wasm-pkg-core/tests/fixtures/cli-example/.cargo/config.toml)). Once those are updated this should work!

lann:
Superseded by #146

PR #146, opened 2025-01-31
Upgrade wasm-tools to 0.224

lann:
- BREAKING: Integrate upstream metadata overhaul. The set of fields available changed, meaning this may break some existing config files.
- Adapt to changed `WitPrinter` interface
- Update targets to `wasm32-wasip1`

### DIFF
--- a/crates/wasm-pkg-core/src/config.rs
+++ b/crates/wasm-pkg-core/src/config.rs
@@ -16,6 +16,7 @@ pub const CONFIG_FILE_NAME: &str = "wkg.toml";
 /// The structure for a wkg.toml configuration file. This file is entirely optional and is used for
 /// overriding and annotating wasm packages.
 #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
 pub struct Config {
     /// Overrides for various packages
     #[serde(default, skip_serializing_if = "Option::is_none")]
@@ -59,6 +60,7 @@ impl Config {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
+#[serde(deny_unknown_fields)]
 pub struct Override {
     /// A path to the package on disk. If this is set, the package will be loaded from the given
     /// path. If this is not set, the package will be loaded from the registry.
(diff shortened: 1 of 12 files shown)


crates/wasm-pkg-core/src/config.rs
```diff
@@ -75,56 +74,22 @@ pub struct Override {
 pub struct Metadata {
     /// The authors of the package.
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub authors: Option<Vec<String>>,
```
thomastaylor312, line 78: I think for one version we should probably keep these, aliasing where it makes sense (e.g. mapping the first item in the `authors` vec to `author`, aliasing the json deserialization for `licenses`) and deprecating the other fields. The latter might not matter as much because the deserializer should just ignore unknown fields.

crates/wkg/tests/fixtures/wasi-http/wkg.toml
thomastaylor312, line 1: Can we add a test with the new fields as well just as a sanity/smoke test that it doesn't fail parsing?

thomastaylor312 (changes requested):
Two discussion items around the breaking changes

crates/wasm-pkg-core/src/config.rs
lann: I'm on the fence here due to a couple of issues:

- There is no equivalent for `documentation` or `categories` so these would be silently discarded
- The URLs are validated as...URLs
- The new wasm-metadata validates license SPDX formatting

Given these basically unavoidable incompatibilities it feels like we might as well rip the band-aid off on this one unless there are existing consumers we know would benefit from a transition.

> the deserializer should just ignore unknown fields.

I wonder if we should change this; accepting e.g. misspelled fields is kind of gnarly UX... :grimacing:

crates/wkg/tests/fixtures/wasi-http/wkg.toml
lann: Sure. I was going with minimal "only fix broken tests" approach but that probably makes sense.

crates/wasm-pkg-core/src/config.rs
lann: I added aliases for `license` and `repository` as those seem harmless; skipped `authors` since the field type changed and there is no canonical OCI way to specify multiple authors. Added `deny_unknown_fields`.

thomastaylor312 (approved):

PR #150, opened 2025-02-13
Add support for OCI authors annotation

yoshuawuyts:
Follow up to #146, this also allows the authors to be read from the wasm binary and written to the OCI package. Thanks!


lann:
Should this wait on https://github.com/bytecodealliance/wasm-tools/pull/2052?

yoshuawuyts:
nah, I don't think so. I think we may want to hold back on populating the `Author` structured in producer toolchains until the next `wasm-tools` release - but I don't think there is any harm in landing it on the publishing side now. We should just make sure to upgrade to the next wasm-tools release when it is next published.

PR #152, opened 2025-03-03
feat!: Removes warg

thomastaylor312:
This PR only focuses on removing warg with some basic updates to the documentation. Ideally we can change the config file as well, but I figured it would be good to start


vados-cosmonic:
Duplicate of the more recent #193

PR #157, opened 2025-03-14
update wasm-tools deps to 0.227

pchickey:
and rename local struct member `author` to `authors` to align with the same change in wasm-metadata

### DIFF
--- a/crates/wasm-pkg-core/src/config.rs
+++ b/crates/wasm-pkg-core/src/config.rs
@@ -75,8 +75,9 @@ pub struct Override {
 #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
 #[serde(deny_unknown_fields)]
 pub struct Metadata {
-    /// The author(s) of the package.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
+    /// The author(s) of the package. Alias supports prior definition as `author`.
+    /// Note that unlike in a Cargo.toml, this authors is a string, not a list of string.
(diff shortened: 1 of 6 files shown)


thomastaylor312:
Looks good but you'll need one more change to the metadata section of the toml manifest so it can handle a vec of authors. That is what caused the test failure here

pchickey:
Thanks. I don't have docker setup to run tests locally so I am letting CI do it for me, and forgetful about checking in.

pchickey:
Flagging this for @yoshuawuyts: the Authors type should be changed to be a list, not an option, upstream https://github.com/bytecodealliance/wasm-pkg-tools/pull/157/files#diff-dd719d29b66d328d4ff372277b6dcfb8ce8b7beaa6dd14a287c054f07c1be7a2

pchickey:
That's unfortunate. Ok! Lets ship it as is then.

crates/wasm-pkg-core/src/config.rs
```diff
@@ -77,7 +77,7 @@ pub struct Override {
 pub struct Metadata {
     /// The author(s) of the package.
     #[serde(default, skip_serializing_if = "Option::is_none")]
```
thomastaylor312, line 79: Can you please add an alias for `author` so the old versions of file still parse correctly?

thomastaylor312 (changes requested):

crates/wasm-pkg-core/src/config.rs
pchickey: Good idea, done

thomastaylor312 (approved):

PR #165, opened 2025-06-19
Reduce level of invalid semver OCI tag to WARN

lann:
WASI packages are publishing cosign signatures with the same prefix as WIT packages: https://github.com/WebAssembly/WASI/pkgs/container/wasi%2Fhttp

Pragmatically this turns this warning into unhelpful noise, so we'll drop the level to debug.

### DIFF
--- a/crates/wasm-pkg-client/src/oci/loader.rs
+++ b/crates/wasm-pkg-client/src/oci/loader.rs
@@ -36,7 +36,7 @@ impl PackageLoader for OciBackend {
                     yanked: false,
                 }),
                 Err(err) => {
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):

thomastaylor312:
Heads up @lann I had a fix for this in #152, but it would probably be better to do this instead: 

```
https://github.com/bytecodealliance/wasm-pkg-tools/pull/152/files#diff-14fca1ad5d1962382cdbc0aba4b858f7ecfbbb4c83bf773f2aebfc266653f418R41
```

I think this is fine too, but probably better to just filter these out entirely

PR #176, opened 2025-10-21
Use registry host rather than URL for creds lookup

itowlson:
There is an issue in GitHub actions where `wkg publish` fails with code `DENIED` even after a successful `docker login`.

The reason appears to be that `wkg` asks `docker-credential` to look up credentials for the schemeful `server_url` (e.g. `https://ghcr.io`), whereas `~/.docker/config.json` contains bare hosts (e.g. `ghcr.io`).  `docker-credential` does not normalise away the schema of the passed-in lookup key, so never matches the schemeful entry.

This PR proposes changing the lookup to use the `oci_registry` of the backend type, rather than the URL.

This also fixes an issue on WSL, where the `docker-credential-desktop.exe` store also requires matching the bare host.

That said, I don't know what _else_ is relying on having the full URL.  The current code has been there for nearly 18 months and I'm assuming it was done that way for a reason.  So if we want to be more conservative, we can try a more complicated solution where we test with the `oci_registry` first, and if that fails, try with the constructed `server_url`.  I'll let wiser heads decide whether that's worth it!

### DIFF
--- a/crates/wasm-pkg-client/src/oci/mod.rs
+++ b/crates/wasm-pkg-client/src/oci/mod.rs
@@ -112,31 +112,17 @@ impl OciBackend {
             ));
         }
 
-        match docker_credential::get_credential(&self.oci_registry) {
-            Ok(DockerCredential::UsernamePassword(username, password)) => {
-                return Ok(RegistryAuth::Basic(username, password));
-            }
-            Ok(DockerCredential::IdentityToken(_)) => {
-                return Err(Error::CredentialError(anyhow::anyhow!(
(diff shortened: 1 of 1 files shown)


itowlson:
@lann I have added checking with scheme if checking without scheme fails.  I'm not able to test with "with scheme" fallback because my creds helper is firmly team "no scheme."  It sounds like you've needed scheme in the past so if you're able to give this a try then that would be super helpful - thanks!

lann:
I've verified no regression in at least one scenario, which I think is good enough given the limited risk here.

PR #179, opened 2025-12-15
feat(core,logging): attach location/dependency where `wasm_pkg_core::wit::resolve_dependencies` failed

mkatychev:
### Summary

`wkg wit build` currently returns an unhelpful report when path resolution fails during `resolve_dependencies`:

```sh-session
$ wkg wit build --wit-dir ./wit --config wkg-config.toml
 INFO wasm_pkg_core::lock: dependency package was not in the lock file 
    package_ref=namespace:pkg1 registry=Some("my-registry.com") requirement==0.1.0
DEBUG wasm_pkg_core::resolver: dependency already exists and override is not set, ignoring
    name=namespace:pkg1
 INFO wasm_pkg_core::lock: dependency package was resolved by the lock file
    package_ref=namespace:pkg2 registry=Some("my-registry.com") requirement==0.1.0 resolved_version=0.1.0
DEBUG wasm_pkg_core::resolver: dependency already exists and override is not set, ignoring
    name=namespace:pkg2
Error: Unable to resolve dependencies

Caused by:
    No such file or directory (os error 2)

```

In the case above, `pkg3` was the cause of the failure with logging failing to show it.

### Changes

* Added `Display` impl for `Dependency`
* `resolve_dependencies` now attaches paths/dependencies to the relevant `anyhow::Error`:
  ```  
  Caused by:
      0: namespace/pkg3
      1: No such file or directory (os error 2)
   ```
* ran `cargo clippy --fix`
* ran `typos ./crates --write-changes`

### DIFF
--- a/crates/wasm-pkg-common/src/config.rs
+++ b/crates/wasm-pkg-common/src/config.rs
@@ -191,27 +191,21 @@ impl Config {
     /// - The default registry
     /// - Hard-coded fallbacks for certain well-known namespaces
     pub fn resolve_registry(&self, package: &PackageRef) -> Option<&Registry> {
+        let namespace = package.namespace();
+        // look in `self.package_registry_overrides `
+        // then in `self.namespace_registries`
         if let Some(reg) = self
             .package_registry_overrides
             .get(package)
+            .or_else(|| self.namespace_registries.get(namespace))
             .map(|pkg| pkg.registry())
         {
             return Some(reg);
-        }
-
-        if let Some(reg) = self
-            .namespace_registries
-            .get(package.namespace())
-            .map(RegistryMapping::get_registry)
-        {
+        } else if let Some(reg) = self.default_registry.as_ref() {
             return Some(reg);
         }
 
-        if let Some(reg) = self.default_registry.as_ref() {
-            return Some(reg);
-        }
-
-        self.fallback_namespace_registries.get(package.namespace())
+        self.fallback_namespace_registries.get(namespace)
     }
 
     /// Returns the default registry.


crates/wasm-pkg-common/src/config.rs
```diff
@@ -181,27 +191,27 @@ impl Config {
...
-        } else {
-            None
+            return Some(reg);
         }
+
+        if let Some(reg) = self.default_registry.as_ref() {
+            return Some(reg);
+        }
+
+        self.fallback_namespace_registries.get(package.namespace())
```
lann: Not a huge problem but the previous `if/else if/else` structure seems like more typical rust code.

lann (commented):
Mostly looks good. Some style and error message suggestions.

crates/wasm-pkg-core/src/wit.rs
```diff
@@ -179,16 +179,16 @@ pub async fn resolve_dependencies(
...
-            let dep = match (ovride.path.as_ref(), ovride.version.as_ref()) {
-                (Some(path), None) => {
-                    let path = tokio::fs::canonicalize(path).await?;
-                    Dependency::Local(path)
-                }
-                (Some(path), Some(_)) => {
-                    tracing::warn!("Ignoring version override for local package");
-                    let path = tokio::fs::canonicalize(path).await?;
+            let dep = match (ovr.path.as_ref(), ovr.version.as_ref()) {
+                (Some(path), v @ None | v @ Some(_)) => {
```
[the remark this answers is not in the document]
mkatychev: ah fair point there

crates/wasm-pkg-common/src/config.rs
mkatychev: For code golf's sake the function could be rewritten as this:

```rust
pub fn resolve_registry(&self, package: &PackageRef) -> Option<&Registry> {
    let namespace = package.namespace();
    // look in `self.package_registry_overrides `
    // then in `self.namespace_registries`
    self.package_registry_overrides
        .get(package)
        .or_else(|| self.namespace_registries.get(namespace))
        .map(|pkg| pkg.registry())
        .or(self.default_registry.as_ref())
        .or_else(|| self.fallback_namespace_registries.get(namespace))
}
```

Issue #99, opened 2024-10-01
Should this repository's CLI be merged with `wasm-tools`?


alexcrichton:
This came up as a question during the BA summit last week and we didn't get to talk about it properly (alas!) so I wanted to raise the issue here. @thomastaylor312 originally expressed interest in possibly merging with `wasm-tools` and @macovedj I believe was also interested (please cc more folks if y'all know of more too!)

Historically I believe I was personally of the stance "let's have the tools be separate" to have a clear line between "this thing needs the network" and "this thing doesn't". Given our proliferation of tools I'm at least personally reversing stance and thinking "well that's not obviously a good idea". In that sense, I'd like to open this up to discuss more about whether this is a good idea, why so, and if so how to do it. From a `wasm-tools` perspective the most difficult part I think is that there's a pretty diverse group of folks interested in it (or at least I sort of assume so, I'm doing my best to sort of personify possible personas interested in the tool moreso than channeling folks I actually know about). 

In the abstract "monorepo good" so I'm at least not here initially going to try to enumerate why the two repos should merge. For me visibility of the tool, coordinated release schedules, and ease of learning for newcomes are huge benefits to earn. Instead here I'm going to focus on the possible downsides of a merge. I want to be clear though in that I don't want to try to create a laundry list of "these must all be fixed" but rather make sure that these downsides are considered. I wouldn't consider it necessary or required to solve every last one.

Broadly speaking though bringing networking functionality to a CLI tool is, in my opinion, no trivial thing to take on. That's a huge step up from what `wasm-tools` does today. All networking nowadays typically comes with TLS/authentication as well, which raises questions like which TLS library is used? How does that affect release binaries? How does that affect local developers. With that as sort of background, here's what I would imagine are possible downsides:

* Loss of portability. Right now `wasm-tools` doesn't have really any system dependencies except `std::process` sorta, so it's pretty easy to compile and run on most systems so long as Rust is supported. Pulling in a dependency on TLS tends to not be quite so portable. For example cross-compiles tend to be harder once TLS is in the mix.
* Bigger builds for developers. In a monorepo-style situation developers that want to "run all the tests" tend to have to build quite a lot. This means that anyone developing `wasm-tools` would have to figure out TLS dependencies, make sure they're all available, and satisfy this new build dependency.
[...]

PR #193, opened 2026-02-20
feat!: Removes warg (rebased on main)

karthik2804:
This PR is the rebased version of #152 to make it easier to review and move it forward.


lann:
Given that we're doing this in large part to ease #99 I wonder if we can just use this branch as the source for that migration, combining this smaller breaking change into that larger breaking change.

vados-cosmonic:
We certainly still want to do this, but given that there are some conflicts, @karthik2804 would you mind reworking this to resolve the conflicts? We're happy to remake the PR as well of course but if you have the cycles would be happy to wait a bit and merge this PR.

karthik2804:
I can take a stab at that.

vados-cosmonic (approved):
LGTM 🚀 

Thanks for taking the time to rework this PR @karthik2804 ! I'll hold off on merging right now just in case you have any last things to change, but otherwise let's get this merged!

karthik2804:
@vados-cosmonic I think it can be merged.

PR #200, opened 2026-04-13
Update README with new section on fallback registries. 

hakilebara:
I struggled to understand how `wkg wit fetch` was falling back to ghcr.io to fetch dependencies, given my `wasm-pkg` config file was empty. I added a section to the README on fallback registries. 

Also, I didn't know about .well-known metadata files and added a Wikipedia link to help future readers like me.

[related conversation on Zulip](https://bytecodealliance.zulipchat.com/#narrow/channel/206238-general/topic/default.20registry.20of.20wkg.20wit.20fetch/near/585262514)

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,14 +142,18 @@ config_file = "/a/path"
 auth = { username = "open", password = "sesame" }
 ```
 
-### Well-known metadata
+### Metadata via Well-known URI (`/.well-known`)
 
-For well-used or public registries, we recommend creating a well-known metadata file that is used by
+For well-used or public registries, we recommend creating a [well-known metadata file](https://en.wikipedia.org/wiki/Well-known_URI) that is used by
 the tool chain to simplify configuration and indicate to a client which protocols and mappings to
 use (although this can be set directly in config as well). The `wkg` tool and libraries expect a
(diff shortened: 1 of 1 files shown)


README.md
vados-cosmonic:
```suggestion
For well-used or public registries, we recommend creating a [`.well-known` metadata file](https://en.wikipedia.org/wiki/Well-known_URI) that is used by
the tool chain to simplify configuration and indicate to a client which protocols and mappings to
use (although this can be set directly in config as well). 

The `wkg` tool and libraries expect a `registry.json` file to be present at a specific location to indicate to the tooling where the components are stored. For example, given a registry `example.com`, then the tooling will attempt to find a `registry.json` file at `https://example.com/.well-known/wasm-pkg/registry.json`. 
```

IMO we can go even further here! Feel free to take or leave these changes :)

README.md
```diff
@@ -144,7 +144,7 @@ auth = { username = "open", password = "sesame" }
 
 ### Well-known metadata
```
vados-cosmonic:
```suggestion
### Metadata via Well-known URI (`/.well-known`)
```

macovedj:
Looks good to me.  Thanks for the contribution!

PR #202, opened 2026-04-24
fix(resolver,config): remove references to root package in override dependencies

mkatychev:
Currently having a "workspace" level `wkg.toml` is not possible since for any package being built, they will also show up as a dependency in the config:

```toml
# wkg.toml
[overrides]
"namespace:foo" = { path = "./foo" }
"namespace:bar" = { path = "./bar" }
```

```sh-session
$ wkg wit build --wit-dir ./foo/

Error: failed to merge package from directory `.`

Caused by:
    attempting to re-add package `namespace:foo@7.2.0` when it's already present in this `Resolve`
```

### DIFF
--- a/crates/wasm-pkg-core/src/resolver.rs
+++ b/crates/wasm-pkg-core/src/resolver.rs
@@ -730,6 +730,11 @@ impl DependencyResolutionMap {
                     package,
                 } => {
                     source_files.extend(package.source_map.source_files().map(Path::to_path_buf));
+                    // handle cases for a workspace level `wkg.toml` where overrides may reference
+                    // the package being built
+                    if package.main.name == root.main.name {
+                        continue;
+                    }
                     merged.push_group(package).with_context(|| {
                         format!(
                             "failed to merge dependency `{name}`",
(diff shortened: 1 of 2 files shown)


crates/wasm-pkg-core/src/resolver.rs
```diff
@@ -730,6 +730,11 @@ impl DependencyResolutionMap {
                     package,
                 } => {
                     source_files.extend(package.source_map.source_files().map(Path::to_path_buf));
+                    // handle cases for a workspace level `wkg.toml` where overrides may reference
+                    // the package being built
+                    if package.main.name == root.main.name {
+                        continue;
+                    }
```
mkatychev, line 737: Not sure if this or perhaps `resolve_dependencies` is the best place to handle this case but this should apply to local or registry packages regardless if the dependency name+namespace+version match that of the root: 
https://github.com/bytecodealliance/wasm-pkg-tools/blob/0d3b316572385cc2ed5eb02069fd55db9de12e18/crates/wasm-pkg-core/src/wit.rs#L180-L212
My one concern is the comment on L180 making the `get_pacakges` move from L212 tricky.

lann:
I'm not following the problem scenario here. Could you sketch it out with a little more detail?

mkatychev:
> I'm not following the problem scenario here. Could you sketch it out with a little more detail?

Hi @lann I've discussed this with @macovedj and there's probably a better approach to solve the issue of package workspacing/development.

PR #209, opened 2026-06-01
wasm-pkg-client: support extra_root_certificates and accept_invalid_certificates when fetching/publishing

### LINKED ISSUE #127: Support `extra_root_certificates` and `accept_invalid_certificates` when fetching/publishing
Often with internal registries there are self-signed certificates in use that are interpreted as invalid. I would love to be able to either specify these flags to ignore invalid certs and supply additional certs to trust at least in the wasm-pkg-client library, and perhaps even with the `wkg` CLI as well.

If there's appetite for this to be contributed, I'm happy to add this feature

ryan-surname-p:
closes #127 

Lifts ClientConfig fields `accept_invalid_certificates` and `extra_root_certificates` into the OciRegistryConfig such that it can be passed through ClientConfig and ultimately to the reqwest client.

### DIFF
--- a/crates/wasm-pkg-client/src/oci/config.rs
+++ b/crates/wasm-pkg-client/src/oci/config.rs
@@ -3,7 +3,7 @@ use base64::{
     engine::{DecodePaddingMode, GeneralPurpose, GeneralPurposeConfig},
     Engine,
 };
-use oci_client::client::ClientConfig;
+use oci_client::client::{Certificate, CertificateEncoding, ClientConfig};
 use secrecy::{ExposeSecret, SecretString};
 use serde::{Deserialize, Serialize, Serializer};
 use wasm_pkg_common::{config::RegistryConfig, Error};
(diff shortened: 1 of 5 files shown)


crates/wasm-pkg-client/src/oci/config.rs
```diff
@@ -161,6 +186,47 @@ fn serialize_secret<S: Serializer>(
...
+#[derive(Clone, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+enum TomlCertificateEncoding {
+    Der,
+    Pem,
+}
+
+#[derive(Clone, Deserialize, Serialize)]
+struct TomlCertificate {
+    encoding: TomlCertificateEncoding,
```
lann, line 198: You can pretty trivially detect format by looking for a prefix of `-----BEGIN` for PEM.

lann: Ah I see you're just taking this from `oci-client`. I guess it would be better to fix this there.

ryan-surname-p: Yep then ultimately to reqwest 

https://github.com/oras-project/rust-oci-client/blob/d446d98cd94ac0f37f6b5b13646d78866b451f7a/src/client.rs#L2238-L2247

lann:
Thanks!

PR #210, opened 2026-06-02
feat: added `--dry-run` to `wkg publish`

mkatychev:
Added ability to dry run publication for the three supported backends attempting to do as many of the present logical steps, stopping short of doing the publication itself.

Added various logging steps for issues I ran into attempting to publish to OCI that were not currently evident.

### DIFF
--- a/crates/wasm-pkg-client/src/warg/publisher.rs
+++ b/crates/wasm-pkg-client/src/warg/publisher.rs
@@ -18,6 +18,7 @@ impl PackagePublisher for WargBackend {
         package: &PackageRef,
         version: &Version,
         data: PublishingSource,
+        dry_run: bool,
     ) -> Result<(), crate::Error> {
         // store the Wasm in Warg cache, so that it is available to Warg client for uploading
         let content = self
@@ -36,6 +37,10 @@ impl PackagePublisher for WargBackend {
 
         // convert package name to Warg package name
         let name = super::package_ref_to_name(package)?;
+
+        if dry_run {
+            return Ok(());
+        }
 
         // start Warg publish, using the keyring to sign
         let version = version.clone();
--- a/crates/wasm-pkg-core/src/config.rs
+++ b/crates/wasm-pkg-core/src/config.rs
@@ -30,6 +30,7 @@ pub struct Config {
 impl Config {
     /// Loads a configuration file from the given path.
     pub async fn load_from_path(path: impl AsRef<Path>) -> Result<Config> {
+        tracing::info!(path = %path.as_ref().display(), "loading wkg config file");
         let contents = tokio::fs::read_to_string(path)
             .await
             .context("unable to load config from file")?;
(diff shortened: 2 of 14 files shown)


crates/wasm-pkg-core/src/config.rs
```diff
@@ -30,6 +30,7 @@ pub struct Config {
 impl Config {
     /// Loads a configuration file from the given path.
     pub async fn load_from_path(path: impl AsRef<Path>) -> Result<Config> {
+        tracing::info!(path = %path.as_ref().display(), "loading wkg config file");
```
mkatychev, line 33: It is unfortunate that the tracing deps has to be added for this line but it was particularly valuable to have visibility in what file actually gets resolved here.

crates/wasm-pkg-core/src/resolver.rs
```diff
@@ -530,7 +530,8 @@ impl<'a> DependencyResolver<'a> {
             } else {
                 let versions =
                     load_package(&mut self.packages, &self.client, dependency.package.clone())
-                        .await?
+                        .await
+                        .with_context(|| format!("package: {}", dependency.package.clone()))?
```
lann, line 534: Looks like a dupe with the next line?

```suggestion
```

crates/wasm-pkg-client/src/warg/publisher.rs
```diff
@@ -37,6 +38,10 @@ impl PackagePublisher for WargBackend {
         // convert package name to Warg package name
         let name = super::package_ref_to_name(package)?;
 
+        if dry_run {
+            return Ok(());
```
lann, line 42: Totally fine if you don't want to make this change but for dry runs I've found it can be nice to model them as an error variant, e.g.


```suggestion
            return Err(Error::DryRun);
```

crates/wasm-pkg-core/src/resolver.rs
mkatychev: I thought so too, but if you look, the next line is actually an option because `load_package` returns `Result<Option<T>>` so the added `with_context` is if the loading logic failed rather than if the package exists (this issue I specifically ran into).

https://github.com/bytecodealliance/wasm-pkg-tools/blob/feba4ca6f8d5a95a403c745ba0a127b547d2973f/crates/wasm-pkg-core/src/resolver.rs#L601-L605

crates/wasm-pkg-client/src/warg/publisher.rs
mkatychev: I'm following `cargo publish --dry-run` logic as much as possible and they seem to throw an exit code of zero.

I also use cargo publish dry runs as CI steps that would abort on exit code so this would cause issues if one wanted to check that a current commit/PR _would_ be able to publish.

crates/wasm-pkg-core/src/resolver.rs
lann: Ah confusing!

crates/wasm-pkg-client/src/warg/publisher.rs
lann: Yeah you would catch the varient down-stack and exit 0. Anyway definitely not a blocker.

lann (approved):
I'll merge whenever you're ready.

mkatychev:
Ready now, @lann. I have a follow-up PR to thins that allows `wkg publish` to point to a directory as well, this should open the way for us to publish multiple packages at once (semi) atomically.

PR #212, opened 2026-06-02
feat(publish): allow directory path for `wkg publish`

mkatychev:
Moved parts of `wkg wit build` into shared function to allow logic reuse between publish and build:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/a84022459b8c3e71f4d8d1304eca71a7c3d55277/crates/wkg/src/wit.rs#L113-L124

`wkg publish ./wit/` will abort if it would result in changing `wkg.lock`

### DIFF
--- a/crates/wasm-pkg-common/src/lib.rs
+++ b/crates/wasm-pkg-common/src/lib.rs
@@ -45,7 +45,7 @@ pub enum Error {
     NoRegistryForNamespace(Label),
     #[error("Package not found")]
     PackageNotFound,
-    #[error("registry error: {0}")]
+    #[error("registry error")]
     RegistryError(#[source] anyhow::Error),
     #[error("registry metadata error: {0:#}")]
     RegistryMetadataError(#[source] anyhow::Error),
--- a/crates/wasm-pkg-common/src/lib.rs
+++ b/crates/wasm-pkg-common/src/lib.rs
@@ -45,7 +45,7 @@ pub enum Error {
     NoRegistryForNamespace(Label),
     #[error("Package not found")]
     PackageNotFound,
-    #[error("registry error")]
+    #[error("registry error: {0:#}")]
     RegistryError(#[source] anyhow::Error),
     #[error("registry metadata error: {0:#}")]
     RegistryMetadataError(#[source] anyhow::Error),
--- a/crates/wasm-pkg-core/src/lock.rs
+++ b/crates/wasm-pkg-core/src/lock.rs
@@ -48,6 +48,9 @@ impl PartialEq for LockFile {
         self.packages == other.packages && self.version == other.version
     }
 }
+/// Compares a [`LockFile`] against a `(version, packages)` snapshot. This is used to
+/// detect whether a build would mutate the lock file (e.g. before publishing) without
+/// having to hold a second [`LockFile`].
 impl PartialEq<(u64, BTreeSet<LockedPackage>)> for LockFile {
     fn eq(&self, other: &(u64, BTreeSet<LockedPackage>)) -> bool {
         self.packages == other.1 && self.version == other.0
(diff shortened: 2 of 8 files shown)


crates/wasm-pkg-core/src/lock.rs
```diff
@@ -48,6 +48,11 @@ impl PartialEq for LockFile {
         self.packages == other.packages && self.version == other.version
     }
 }
```
ricochet, line 50:
```suggestion
}
/// Compares a [`LockFile`] against a `(version, packages)` snapshot. This is used to
/// detect whether a build would mutate the lock file (e.g. before publishing) without
/// having to hold a second [`LockFile`].
```

crates/wkg/src/main.rs
```diff
@@ -260,10 +262,49 @@ impl PublishArgs {
...
+                    "wkg.lock would be updated during publish, aborting"
+                ))
+                .with_context(|| {
+                    format!(
+                        "Run `wkg wit build {}` before attempting to publish",
+                        self.path.to_string_lossy()
+                    )
+                });
+            }
+
```
ricochet, line 289:
```suggestion

            // Sanitize the package ref for use as a filename prefix: `namespace:name`
            // contains characters (`:`, `/`) that are invalid in filenames on some
            // platforms (notably Windows).
            let prefix: String = build_ref.to_string().replace([':', '/'], "_");
```

crates/wkg/src/main.rs
```diff
@@ -260,10 +262,49 @@ impl PublishArgs {
...
+            let tmp = tempfile::Builder::new()
+                .prefix(&build_ref.to_string())
```
ricochet:
```suggestion
                .prefix(&prefix)
```

crates/wasm-pkg-common/src/lib.rs
```diff
@@ -45,7 +45,7 @@ pub enum Error {
     NoRegistryForNamespace(Label),
     #[error("Package not found")]
     PackageNotFound,
-    #[error("registry error: {0}")]
+    #[error("registry error")]
```
ricochet:
```suggestion
    #[error("registry error: {0:#}")]
```
We still want the full chain and to be consistent with RegistryMetadataError.

crates/wkg/src/main.rs
mkatychev: Good catch, should probably handle `/` for https://github.com/bytecodealliance/wasm-pkg-tools/pull/214 as well

crates/wkg/src/main.rs
```diff
@@ -250,20 +254,50 @@ impl PublishArgs {
...
+        // file first. _tmp is held until the publish completes so the file
+        // isn't deleted out from under us.
+        let (publish_path, _tmp) = if self.path.is_dir() {
+            let mut lock_file = LockFile::load(true).await?;
+            let prev_lock_ref = (lock_file.version, lock_file.packages.clone());
+            let (pkg_ref, _, bytes) =
+                wit::build_wit_dir(&self.path, client.clone(), &mut lock_file).await?;
+            // There is no way to check if we are in a git repository unlike `cargo publish --allow-dirty` so
+            // check against previous values.
+            if lock_file != prev_lock_ref && !self.dry_run {
```
mkatychev, line 278: I'm not sure this check is even needed yet

vados-cosmonic (approved):
Hey @mkatychev this looks great! 

Either in this PR or as a follow up (this PR has certainly been waiting too long!), what do you think about adding a regression test for this real quick? If there are any convenient-to-add adjacent tests those would be great too, but reading over the code this makes me think that we should really be testing the at multiple levels (core/client & CLI) to make sure we don't regress.

ricochet (approved):

mkatychev:
@vados-cosmonic I agree, one of my hesitations in doing config/data driven coverage testing is that I plan to add overlay support cargo style sometime after https://github.com/bytecodealliance/wasm-pkg-tools/pull/215. Note this commit from the multi package publishing PR:

https://github.com/mkatychev/wasm-pkg-tools/blob/9de6af4ed5fad54b71f7ccb8b8e77b953e0910c5/crates/wkg/src/main.rs#L273-L304

PR #213, opened 2026-06-03
wasm-pkg-client: validate semver compliance on publish

### LINKED ISSUE #128: Package publishing should check semver compat
The `wit-component` crate provides a function [`semver_check`](https://docs.rs/wit-component/latest/wit_component/fn.semver_check.html) for checking that two packages are semver compatible.

The publish path should fetch any packages that are in the same semver series, and then check to make sure the to-be-published package is semver compatible with them.

We could allow disabling this check with an optional `--no-verify` flag for exceptional situations, but I believe this behavior should be enforced by default.

ryan-surname-p:
closes #128 

Summary 

By default publish: 

- rejects duplicate registrations
- ensures component is compatible with it's nearest server series neighbors using [wit_component::check_semver](https://docs.rs/wit-component/latest/wit_component/fn.semver_check.html)

The changeset is fairly large I know though a material amount of the new code are tests and moving functions around to keep things readable.

### DIFF
--- a/crates/wasm-pkg-client/src/decoded_component.rs
+++ b/crates/wasm-pkg-client/src/decoded_component.rs
@@ -1,0 +1,196 @@
+use crate::{ContentStream, PublishingSource};
+use futures_util::TryStreamExt;
+use std::io::Read;
+use tokio::io::AsyncSeekExt;
+use tokio_util::io::{StreamReader, SyncIoBridge};
+use wasm_pkg_common::{
+    package::{PackageRef, Version},
+    Error,
+};
+use wit_component::DecodedWasm;
+
+pub struct DecodedComponent {
+    version: Version,
+    package_ref: PackageRef,
+    decoded_wasm: DecodedWasm,
+}
+
+impl DecodedComponent {
+    pub async fn from_publishing_source(
+        data: PublishingSource,
+    ) -> Result<(PublishingSource, DecodedComponent), Error> {
+        let (reader, decoded_wasm) = decode(SyncIoBridge::new(data)).await?;
+        let (package_ref, version) = extract_package_version(&decoded_wasm)?;
+
+        let mut data = reader.into_inner();
+        data.rewind().await?;
+
+        Ok((
+            data,
+            DecodedComponent {
+                version,
+                package_ref,
+                decoded_wasm,
+            },
+        ))
+    }
+
+    /// Like [`Self::from_publishing_source`] but overrides the derived
+    /// `(package, version)` identity with `package_override` when supplied.
+    pub async fn from_publishing_source_with_package(
+        data: PublishingSource,
+        package_override: Option<(PackageRef, Version)>,
+    ) -> Result<(PublishingSource, DecodedComponent), Error> {
+        let (data, mut decoded) = Self::from_publishing_source(data).await?;
+        if let Some((p, v)) = package_override {
+            decoded.package_ref = p;
+            decoded.version = v;
+        }
+        Ok((data, decoded))
+    }
+
+    /// Construct from a registry content stream. Callers already know the
+    /// `(package, version)` identity from the registry listing they followed
+    /// to get here, so we take it as input rather than re-deriving it from
+    /// the wasm metadata.
+    pub async fn from_content_stream(
+        stream: ContentStream,
+        package_ref: PackageRef,
+        version: Version,
+    ) -> Result<DecodedComponent, Error> {
+        let reader = SyncIoBridge::new(StreamReader::new(stream.map_err(std::io::Error::other)));
+        let (_reader, decoded_wasm) = decode(reader).await?;
+        Ok(DecodedComponent {
+            version,
+            package_ref,
+            decoded_wasm,
+        })
+    }
+
+    pub fn version(&self) -> &Version {
+        &self.version
+    }
+
+    pub fn package(&self) -> &PackageRef {
+        &self.package_ref
+    }
+
+    /// Check that `self` and `other` are semver-compatible neighbors in the
+    /// same cargo-`^` compatibility range.
+    pub fn semver_check(&self, other: &DecodedComponent) -> Result<(), Error> {
+        // `wit_component::semver_check` is asymmetric: its `new` may add
+        // imports / drop exports relative to its `prev`. To get a symmetric
+        // additive-only gate between two published versions we pass the
+        // newer-in-time release as `prev` and the older as `new`.
+        let (older, newer) = if self.version < other.version {
+            (self, other)
+        } else {
+            (other, self)
+        };
+
+        let (prev_resolve, prev_world) = extract_resolve_and_world_id(&newer.decoded_wasm)?;
+        let (new_resolve, new_world) = extract_resolve_and_world_id(&older.decoded_wasm)?;
+
+        // Merge resolves, remap merged resolve, check for incompatibility
+        let mut merged = prev_resolve.clone();
+        let new_world = merged
+            .merge(new_resolve.clone())
+            .and_then(|remap| remap.map_world(new_world, None))
+            .map_err(Error::InvalidComponent)?;
+
+        wit_component::semver_check(merged, prev_world, new_world).map_err(|e| {
+            Error::SemverIncompatible {
+                previous: older.version.clone(),
+                new: newer.version.clone(),
+                source: e,
+            }
+        })
+    }
+}
+
+async fn decode<R>(reader: R) -> Result<(R, DecodedWasm), Error>
+where
+    R: Read + Send + 'static,
+{
+    // wit_component::decode_reader is CPU-bound sync work
+    // run it on the blocking pool so we don't stall an async worker thread
+    // see also: https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code
+    tokio::task::spawn_blocking(move || {
+        let mut reader = reader;
+        let decoded_wasm =
+            wit_component::decode_reader(&mut reader).map_err(Error::InvalidComponent)?;
+        Ok::<_, Error>((reader, decoded_wasm))
+    })
+    .await
+    .map_err(|e| Error::IoError(std::io::Error::other(e)))?
+}
+
+/// Extract the package name and version from a decoded candidate.
+fn extract_package_version(decoded: &DecodedWasm) -> Result<(PackageRef, Version), Error> {
+    let resolve = decoded.resolve();
+    let package_id = match decoded {
+        wit_component::DecodedWasm::Component(_, world_id) => {
+            resolve.worlds[*world_id].package.ok_or_else(|| {
+                crate::Error::InvalidComponent(anyhow::anyhow!(
+                    "component world or package not found"
+                ))
+            })?
+        }
+        wit_component::DecodedWasm::WitPackage(_, pkg) => *pkg,
+    };
+    let (package, version) = resolve
+        .package_names
+        .iter()
+        .find_map(|(pkg, id)| {
+            // SAFETY: We just parsed this from wit and should be able to unwrap. If it
+            // isn't a valid identifier, something else is majorly wrong
+            (*id == package_id).then(|| {
+                (
+                    PackageRef::new(
+                        pkg.namespace.clone().try_into().unwrap(),
+                        pkg.name.clone().try_into().unwrap(),
+                    ),
+                    pkg.version.clone(),
+                )
+            })
+        })
+        .ok_or_else(|| {
+            crate::Error::InvalidComponent(anyhow::anyhow!(
+                "component package {package_id:?} not found"
+            ))
+        })?;
+
+    let version = version.ok_or_else(|| {
+        crate::Error::InvalidComponent(anyhow::anyhow!(
+            "component package version not found in the Wasm binary\n\
+            \n\
+            The Wasm file was built without a version in the WIT `package` statement.\n\
+            Add a version to the `package` statement in your .wit file, e.g.:\n\
+            \n\
+            \tpackage example:my-package@1.0.0;\n\
+            \n\
+            Alternatively, specify the package and version explicitly with the --package flag:\n\
+            \n\
+            \twkg publish <file> --package <namespace>:<name>@<version>"
+        ))
+    })?;
+    Ok((package, version))
+}
+
+/// Borrow the inner `wit_parser::Resolve` and resolve a concrete `WorldId`.
+/// For a decoded component the world is fixed; for a WIT package we ask
+/// `Resolve::select_world` to pick one — deferred until needed so a
+/// multi-world WIT package can publish its first version unambiguously.
+fn extract_resolve_and_world_id(
+    decoded: &DecodedWasm,
+) -> Result<(&wit_parser::Resolve, wit_parser::WorldId), Error> {
+    match decoded {
+        DecodedWasm::Component(resolve, world_id) => Ok((resolve, *world_id)),
+        DecodedWasm::WitPackage(resolve, pkg) => {
+            let world_id = resolve
+                .select_world(&[*pkg], None)
+                .map_err(Error::InvalidPackage)?;
+            Ok((resolve, world_id))
+        }
+    }
+}
--- a/crates/wasm-pkg-client/src/decoded_component.rs
+++ b/crates/wasm-pkg-client/src/decoded_component.rs
@@ -4,8 +4,8 @@ use std::io::Read;
 use tokio::io::AsyncSeekExt;
 use tokio_util::io::{StreamReader, SyncIoBridge};
 use wasm_pkg_common::{
-    package::{PackageRef, Version},
     Error,
+    package::{PackageRef, Version},
 };
 use wit_component::DecodedWasm;
 
@@ -105,6 +105,26 @@ impl DecodedComponent {
                 source: e,
             }
         })
+    }
+}
+
+impl PartialEq for DecodedComponent {
+    fn eq(&self, other: &Self) -> bool {
+        self.package_ref == other.package_ref && self.version == other.version
+    }
+}
+
+impl Eq for DecodedComponent {}
+
+impl PartialOrd for DecodedComponent {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for DecodedComponent {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        (&self.package_ref, &self.version).cmp(&(&other.package_ref, &other.version))
     }
 }
 
--- a/crates/wasm-pkg-client/src/lib.rs
+++ b/crates/wasm-pkg-client/src/lib.rs
@@ -27,6 +27,7 @@
 //! ```
 
 pub mod caching;
+mod decoded_component;
 mod loader;
 pub mod local;
 pub mod metadata;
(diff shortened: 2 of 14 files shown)


crates/wasm-pkg-client/src/decoded_component.rs
```diff
@@ -0,0 +1,191 @@
...
+                    PackageRef::new(
+                        pkg.namespace.clone().try_into().unwrap(),
+                        pkg.name.clone().try_into().unwrap(),
+                    ),
+                    pkg.version.clone(),
+                )
+            })
+        })
+        .ok_or_else(|| {
+            crate::Error::InvalidComponent(anyhow::anyhow!("component package not found"))
```
mkatychev: could we add the id as part of the error message?

vados-cosmonic:
Hey @ryan-surname-p it looks like some conflicts creeped in -- apologies for the inconvenience but would you mind resolving those? Then we can get this merged.

ryan-surname-p:
No problem at all. 

@mkatychev actually reached out to me about this and some work they have in flight as well.

https://github.com/bytecodealliance/wasm-pkg-tools/pull/215

If I understood their request correctly we'd like for that to land first and then work some version of this onto that.

This request seemed reasonable enough to me so the current plan is to monitor the progress of 215 and then adjust accordingly.

mkatychev:
I'd like to hear @vados-cosmonic's thoughts on which should land first.

I suppose one of us will be handling gnarly merge conflicts regardless so it may be fairer for @ryan-surname-p's changes to land first.

ryan-surname-p:
FWIW I'm happy enough to handle either approach I just wanted to make sure everyone was working with the same set of assumptions.

crates/wasm-pkg-client/src/loader.rs
```diff
@@ -31,3 +62,191 @@ pub trait PackageLoader: Send {
...
+
+    /// `list_matching_versions` is a generic `VersionReq` filter; this test
+    /// exercises it with the cargo-`^` shaped series masks that the publish
+    /// gate constructs in `fetch_semver_series` (see `lib.rs`):
+    ///
+    /// - `X.y.z` (X >= 1) -> `X.*`
+    /// - `0.Y.z` (Y >= 1) -> `0.Y.*`
+    /// - `0.0.Z`          -> exact `0.0.Z`
+    #[tokio::test]
+    async fn list_matching_versions_filters_by_version_req_table_driven() {
```
mkatychev:
```suggestion
    #[test]
    fn list_matching_versions_filters_by_version_req_table_driven() {
```

crates/wasm-pkg-client/src/loader.rs
```diff
@@ -31,3 +62,191 @@ pub trait PackageLoader: Send {
...
+        // These cases include the examples from the function docs and edge
+        // cases for lane filtering behavior.
+        let cases = [
+            Case {
+                name: "target 0.0.0 -> 0.0.*",
+                req: "~0.0.*",
+                sort: VersionSort::Ascending,
+                history: &["0.0.0", "0.0.1", "0.1.0", "1.0.0"],
+                expected: &["0.0.0", "0.0.1"],
+            },
```
mkatychev: These tests look very go inspired, pretty nice.

mkatychev: Now that I think about it, it may be better to reuse [`rstest`](https://docs.rs/rstest/latest/rstest/) for cases and fixtures:
https://github.com/bytecodealliance/wasm-pkg-tools//blob/9dc967c687c6a9ee34c719c5383dfa71ce0cdbc6/crates/wasm-pkg-core/tests/fetch.rs#L13-L21

We could still use the `Case`s defined and destructure them: https://docs.rs/rstest/latest/rstest/attr.rstest.html#destructuring-inputs

crates/wkg/src/main.rs
```diff
@@ -246,6 +246,10 @@ struct PublishArgs {
     #[arg(long)]
     dry_run: bool,
 
+    /// Disable semver compatibility checks.
+    #[arg(long)]
+    no_verify: bool,
```
mkatychev: `skip_semver_check` is better than `no_verify` as an unambiguous flag IMO.
```suggestion
    skip_semver_check: bool,
```

crates/wasm-pkg-client/src/loader.rs
```diff
@@ -1,19 +1,50 @@
...
+
+        match sort {
+            VersionSort::Ascending => versions.sort_by(|a, b| a.version.cmp(&b.version)),
+            VersionSort::Descending => versions.sort_by(|a, b| b.version.cmp(&a.version)),
+        };
+
+        let matching: Vec<VersionInfo> = versions
+            .into_iter()
+            .filter(|v| predicate.matches(&v.version))
+            .collect();
```
mkatychev, line 39: should we move the filter before the `sort_by` calls so there is less to sort?

crates/wasm-pkg-client/src/loader.rs
```diff
@@ -31,3 +62,191 @@ pub trait PackageLoader: Send {
...
+        input.parse().expect("valid semver in test case")
+    }
+
+    fn versions(inputs: &[&str]) -> Vec<Version> {
+        inputs.iter().map(|s| v(s)).collect()
+    }
+
+    /// `list_matching_versions` is a generic `VersionReq` filter; this test
+    /// exercises it with the cargo-`^` shaped series masks that the publish
+    /// gate constructs in `fetch_semver_series` (see `lib.rs`):
```
mkatychev: Is there a reason we have doc comments here?

mkatychev (changes requested):

crates/wasm-pkg-client/src/decoded_component.rs
```diff
@@ -0,0 +1,193 @@
...
+{
+    tokio::task::spawn_blocking(move || {
+        let mut reader = reader;
+        let decoded_wasm =
+            wit_component::decode_reader(&mut reader).map_err(Error::InvalidComponent)?;
+        Ok::<_, Error>((reader, decoded_wasm))
+    })
+    .await
+    .map_err(|e| Error::IoError(std::io::Error::other(e)))?
+}
```
mkatychev, line 146: Does this function need to produce a future? If we removed the outer `spawn_blocking`, would we be able to remove the `async`?

ryan-surname-p: This is likely just my inexperience with rust async showing but the idea was to enable concurrent package decoding. 

https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html

If this is too much though I can back it out.

crates/wasm-pkg-client/src/loader.rs
ryan-surname-p: Yes we should

crates/wasm-pkg-client/src/loader.rs
ryan-surname-p: yes, up until relatively recently I was primarily working with go. 

Thanks for pointing out rstest

crates/wkg/src/main.rs
ryan-surname-p: agreed, thank you for the suggestion

crates/wasm-pkg-client/src/loader.rs
ryan-surname-p: The function we are testing, `PackageLoader::list_matching_versions`, invokes `PackageLoader::list_all_versions` which is async. 

This is the reason for the use of both `async` and `[tokio::test]` here.

crates/wasm-pkg-client/src/loader.rs
ryan-surname-p: Nope, if memory serves I actually copy pasted this from another source to help guide me when writing test cases.

mkatychev:
@ryan-surname-p the multi publish package PR has landed, could you address the conflicts and https://github.com/bytecodealliance/wasm-pkg-tools/pull/213#discussion_r3485074021?

ryan-surname-p:
Thanks for the heads up, I'll get to this first thing tomorrow. 

I followed up on https://github.com/bytecodealliance/wasm-pkg-tools/pull/213#discussion_r3485074021 but I realize now you weren't actually able to see it because I never submitted my review 🤦

crates/wasm-pkg-client/src/decoded_component.rs
mkatychev: Oh! `spawn_blocking` will spawn a future that [is _not_ async](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html#:~:text=Be%20aware%20that%20tasks%20spawned,prevent%20the%20task%20from%20starting%2E).

What you want to use is a variant of `task::spawn`, which produces futures, and join said futures together; some options:
* https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.join_all
* https://docs.rs/futures-util/latest/futures_util/macro.try_join.html

So basically:
1. you should collect a list of futures created by repeated calls of decode
2. join the one or more futures together
3. `.await` your future list

Basically anytime you `.await`, you poll for that future to resolve which is does not block the thread but does block the current task you are in.

@vados-cosmonic let me know if I missed something/got something wrong.

ryan-surname-p: > Oh! spawn_blocking will spawn a future that is not async.

Yep were are using `spawn_blocking` to obtain a `JoinHandle` for a synchronous operation. 

What makes `spawn_blocking` special is how the task behind that `JoinHandle` is [scheduled](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code); it moves the work to a thread pool designed to be blocked.

If we ran the same synchronous, non-yielding decode on an async worker thread using either `tokio::spawn`, or by  awaiting an [async wrapper](https://github.com/bytecodealliance/wasm-pkg-tools/pull/213/changes#diff-f8e08c622fbe4805ba2bff616f9c562e40830e3789345a16c236345468c1a034R62), the worker would be pinned until decode returns.

The consequence of this would be that no other future scheduled on that worker could make progress in the meantime.

ryan-surname-p: > 1. you should collect a list of futures created by repeated calls of decode
> 2. join the one or more futures together
> 3. .await your future list

If we [walk backwards from decode](https://github.com/bytecodealliance/wasm-pkg-tools/pull/213/changes#diff-7f20b5f5277cdb211982148d419be1e9ef7397f7c7bc7809779c81b3e758b549R208-R216) we will find this is actually what's going on albeit using [futures_concurrency](https://github.com/bytecodealliance/wasm-pkg-tools/pull/213/changes#diff-7f20b5f5277cdb211982148d419be1e9ef7397f7c7bc7809779c81b3e758b549R49) instead of `futures_util`. 

edit: When rebasing I removed `futures_concurrency` and used futures_util instead; thanks for pointing that out!

mkatychev: > What makes `spawn_blocking` special is how the task behind that `JoinHandle` is [scheduled](https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code); it moves the work to a thread pool designed to be blocked.

Thanks for the explain, could we add this as a comment above the `spawn_blocking` call?
EDIT: missed comment there, thanks!

crates/wkg/tests/e2e.rs
```diff
@@ -146,10 +146,12 @@ async fn publish_multiple_transitive_local_packages() {
...
-        std::assert_matches!(
-            &versions[..],
-            [VersionInfo { version, .. }] if version == &expected_version,
-            "{name} should have exactly one published version",
+        assert!(
+            matches!(
+                &versions[..],
+                [VersionInfo { version, .. }] if version == &expected_version,
+            ),
+            "{name} should have exactly one published version, got {versions:?}",
```
mkatychev, line 154: was there something preventing us from using the `assert_matches!` macro? the inner [`pat`](https://lukaswirth.dev/tlborm/decl-macros/minutiae/fragment-specifiers.html#pat) fragment looks the same

mkatychev (commented):
This is looking great, I will take another look tomorrow (thanks for patience!), once we address a few more nits this should be good to merge!

crates/wasm-pkg-client/src/lib.rs
```diff
@@ -169,24 +169,67 @@ impl Client {
...
+                futures_util::future::join_all(prepare_neighbor_ops)
+                    .await
+                    .into_iter()
+                    .collect::<Result<_, _>>()?;
+
+            // verify candidate is in compliance with its semver neighbors
+            if !semver_series.is_empty() {
+                semver_series.push(candidate);
+
+                semver_series.sort_by(|a, b| a.version().cmp(b.version()));
```
mkatychev: Since we're mostly doing comparisons here and the one below by version:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/4e10a3a22a8a7ca82c87ef2352633bd13ba18303/crates/wasm-pkg-client/src/loader.rs#L41-L42

could we do an `impl PartialOrd for DecodedComponent`?
I wonder if we can use [`std::cmp::Reverse`](https://doc.rust-lang.org/std/cmp/struct.Reverse.html) for `VersionSort::Descending`

crates/wkg/src/main.rs
```diff
@@ -326,6 +330,7 @@ impl PublishArgs {
                                 registry: Some(local_registry.clone()),
                                 // we want to publish to "tmp_local_publish" regardless of flags passed in
                                 dry_run: false,
+                                skip_semver_check: true,
```
mkatychev: Would we not want to have have `skip_semver_check` be the same as the `PublishOpts` value?

crates/wasm-pkg-client/tests/publish_semver_check.rs
```diff
@@ -0,0 +1,222 @@
...
+    let config = Config::from_toml(&toml).expect("local-backend config should parse");
+    Client::new(config)
+}
+
+fn wit_for(package: &str, version: &str, shape: Shape) -> String {
+    let world_body = match shape {
+        Shape::Base => "export run: func() -> u32;",
+        Shape::Compatible => "export run: func() -> u32;\n    export extra: func() -> u32;",
+        Shape::Incompatible => "export run: func() -> string;",
+    };
```
mkatychev: This match arm made the entire PR make more sense after rereading it (we're adding an extra world), could we make the additional export a bit more obvious or add a comment saying something like "changing the world signature is a non-hotfix change"?
```suggestion
        Shape::Base => "export run: func() -> u32;",
        Shape::Compatible => "export run: func() -> u32;
                              export extra: func() -> u32;",
        Shape::Incompatible => "export run: func() -> string;",
    };
```

mkatychev: Actually, could we make this match arm an `impl Display for Shape`?

crates/wasm-pkg-client/tests/publish_semver_check.rs
```diff
@@ -0,0 +1,222 @@
...
+use wasm_pkg_common::Error;
+
+const NAMESPACE: &str = "example";
+
+#[derive(Clone, Copy)]
+enum Shape {
+    Base,
+    Compatible,
+    Incompatible,
+}
```
mkatychev, line 20: Could we rename this enum to map it to the input it creates rather than the expected behaviour? 

```suggestion
enum WorldDiff {
    AddBase,      // + export base: func() -> u32
    AddExtra,     // + export extra: func() -> u32
    ChangeBase,   // - export base: func() -> u32
                  // + export base: func() -> string
}
```
This could also allow us to create tests cases in the future where we can reason about the changes like a git commit:

```diff
// ChangeBase
world the-world {
-  export base: func() -> u32
+  export base: func() -> string
}
```
EDIT: and perhaps make the extra export actually called `extra` to make the correlation more obvious for individuals of the obtuse proclivity such as myself 🧐?

mkatychev:
`wit_component::check_semver` is really cool, like `cargo-semver-checks` for wit!

If we can address||change the nits about `Shape`, `skip_semver_check`, `impl PartialOrd for DecodedComponent`, and `assert_matches!` then this LGTM!

crates/wasm-pkg-client/tests/publish_semver_check.rs
```diff
@@ -0,0 +1,222 @@
...
+)]
+#[tokio::test]
+// `package` is unique per row -> isolated series. `initial` is `None` to skip
+// seeding, in which case the candidate is the first publish in the series.
+async fn publish_semver_check(
+    #[case] package: &str,
+    #[case] initial: Option<(&str, Shape)>,
+    #[case] candidate: (&str, Shape),
+    #[case] skip_semver_check: bool,
+    #[case] expected: Expected,
```
mkatychev: Is there something stoping us from making expected a `Result`?

if we pass in a `Result` then `assert_eq!` should render the expected/actual diff for us.
```suggestion
    #[case] expected: Result<(), Error>,
```

crates/wkg/tests/e2e.rs
ryan-surname-p: Its inclusion in std does not  appear to be stable across versions; possibly because of this [issue](https://github.com/rust-lang/rust/pull/154999)?


For instance if I try to run the test on my machine I get the following:

```bash
cargo --version
cargo 1.94.1 (29ea6fb6a 2026-03-24)
```

```bash
cargo test
   Compiling wkg v0.16.0 (/Users/r.pridgeon/workspace/wkg/wasm-pkg-tools/crates/wkg)
error[E0433]: failed to resolve: could not find `assert_matches` in `std`
   --> crates/wkg/tests/e2e.rs:149:14
    |
149 |         std::assert_matches!(
    |              ^^^^^^^^^^^^^^ could not find `assert_matches` in `std`

For more information about this error, try `rustc --explain E0433`.
error: could not compile `wkg` (test "e2e") due to 1 previous error
```

Granted I should update my toolchain but the above approach is more portable I believe.

mkatychev: @ryan-surname-p one of the issues is that I think we are running 2021 edition as well:
https://github.com/bytecodealliance/wasm-pkg-tools/pull/220#discussion_r3510620863

Let me see if I can get just that commit merged

crates/wkg/src/main.rs
ryan-surname-p: This is the ephemeral registry used for staging our package so there won't be any other versions in the registry to compare against. 

https://github.com/bytecodealliance/wasm-pkg-tools/blob/main/crates/wkg/src/main.rs#L303-L304

This, as I understand it, is the same reason we hardcode dry_run to false.

mkatychev: the dry run was set to false because:
* we want to _not skip_ doing a full publish to the ephemeral registry
Reading the flag here it says to me: 
* we want to _skip_ checking semver validity when doing a full pubilsh to the ephemeral registry

crates/wkg/tests/e2e.rs
ryan-surname-p: This still fails to compile with edition 2024, updating my toolchain to 1.96.1 however resolves it. 

If we revert this should we also set the MSRV to 1.96.1?

mkatychev: Or use the old compat, I have a `rust-toolchain.toml` in https://github.com/bytecodealliance/wasm-pkg-tools/pull/222. Go with what you think is best.

mkatychev (approved):
LGTM! Excited to use this.

crates/wkg/tests/e2e.rs
ryan-surname-p: If it's all the same to you I'll just leave it as is for now and if it really gets too bothersome we can fix forward later.

PR #215, opened 2026-06-10
feat(client,cli): add multi-packaging support for `wkg publish`

mkatychev:
Added the ability to publish multiple WIT packages at once.
Introduced the `PublishPlan` struct to resolve local references into a sequence of batches for publication:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/40d2dd3f84e0fb46661feb965d2269b2080fb90a/crates/wasm-pkg-core/src/resolver.rs#L846-L853

Reference implementation (cargo multi-package publishing):
* https://www.tweag.io/blog/2025-07-10-cargo-package-workspace/
* https://github.com/rust-lang/cargo/pull/14433/changes


crates/wasm-pkg-client/src/lib.rs
```diff
@@ -191,111 +221,118 @@ impl Client {
...
+                                if custom.registry == registry =>
+                            {
+                                Some(custom.metadata.clone())
+                            }
+                            (RegistryMapping::Custom(custom), false) => {
+                                Some(custom.metadata.clone())
+                            }
+                            _ => None,
+                        }
+                    })
```
mkatychev, line 255: this part should go away once proper overlay support is added.

mkatychev:
Note that this PR is not UX complete, there are two follow-ups to this PR to ensure the feature is intuitive:
* multi package `wkg wit fetch`
* pacakge groups/workspace members to ensure consistency in grouping and avoid calling the same arguments repeatedly

crates/wasm-pkg-client/src/lib.rs
```diff
@@ -191,111 +194,119 @@ impl Client {
...
+        tracing::debug!(?backend_type, "Resolved backend type");
+
+        let source: InnerClient = match backend_type {
+            LOCAL_PROTOCOL => Box::new(LocalBackend::new(registry_config)?),
+            OCI_PROTOCOL => Box::new(OciBackend::new(
+                &registry,
+                &registry_config,
+                &registry_meta,
+            )?),
+            WARG_PROTOCOL => {
```
vados-cosmonic, line 294: Just a note but we should probably remove warg in the next breaking-change release -- I think this was scheduled before but we never got around to doing.

crates/wasm-pkg-client/src/local.rs
```diff
@@ -27,13 +29,25 @@ use crate::{
...
 }
 
+impl LocalConfig {
+    /// Creates a [`Self`] using a generated temporary directory. Upon dropping the [`TempDir`] handle,
+    /// the temporary directory will be deleted.
+    pub fn temp_dir() -> Result<(Self, TempDir), Error> {
+        let handle = TempDir::new()?;
+        let root = handle.path().to_owned();
+        tracing::debug!(registry_dir=%root.display(), "created temporary directory");
+        Ok((Self { root }, handle))
```
vados-cosmonic: Should the `LocalConfig` not hold on to the Temp dir? If some caller gets the `LocalConfig` and the `TempDir` and instantly drops the dir, then the `LocalConfig`'s root path would become invalid, no?

crates/wasm-pkg-client/src/local.rs
```diff
@@ -140,7 +154,7 @@ impl PackagePublisher for LocalBackend {
         let mut out = tokio::fs::File::create(&path)
             .await
             .map_err(|e| registry_path_context(e, &path))?;
-        tracing::info!("publishing to {}", path.display());
+        tracing::debug!("publishing to {}", path.display());
```
vados-cosmonic: NIT: IMO `info` was the right level  for this message.... any super strong reason for debug? I don't feel strongly about this at all, but maybe I can learn something here

crates/wasm-pkg-common/src/package.rs
```diff
@@ -68,14 +68,41 @@ impl FromStr for PackageRef {
...
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
 pub struct PackageSpec {
     pub package: PackageRef,
     pub version: Option<Version>,
 }
 
+impl PackageSpec {
+    // Returns a copy of the associated package reference.
+    pub fn package(&self) -> PackageRef {
```
vados-cosmonic: NIT: I'm not sure we need these getters given that `package` and `version` are already pub?

crates/wasm-pkg-common/src/registry.rs
```diff
@@ -55,3 +56,6 @@ impl TryFrom<String> for Registry {
         Ok(Self(value.try_into()?))
     }
 }
+
+/// Graph of publishable packages with the [`petgraph::Direction`] edges describing the dependency direction.
+pub type DependencyGraph<N> = Acyclic<DiGraph<N, petgraph::Direction>>;
```
vados-cosmonic: Is there any value in avoiding tying ourselves to `petgraph` in our public interface here? I wonder if we should have a `#[non_exhaustive]` struct here that wraps the `Acyclic<Digraph<...>>` rather than directly exposing it.

crates/wasm-pkg-common/Cargo.toml
```diff
@@ -31,6 +31,7 @@ tokio = { workspace = true, optional = true, features = ["fs"] }
 toml = { workspace = true, optional = true }
 thiserror = { workspace = true }
 tracing.workspace = true
+petgraph.workspace = true
```
vados-cosmonic:
```suggestion
petgraph = { workspace = true }
```

let's be consistent... no idea why the tracing linea bove is suddenly `.workspace = true`

crates/wasm-pkg-core/src/resolver.rs
```diff
@@ -834,3 +839,199 @@ fn visit<'a>(
 
     Ok(())
 }
+
+/// Mapping of [`PackageRef`]s to the respective index inside the dependency graph.
+pub type LocalPackageIndex = HashMap<PackageRef, (NodeIndex, PathBuf)>;
+
+/// State for tracking dependencies during upload.
+pub struct PublishPlan {
```
vados-cosmonic, line 850: Does this need to be pub? Could we get away with `pub(crate)` here and other places?

crates/wasm-pkg-core/src/resolver.rs
```diff
@@ -834,3 +839,199 @@ fn visit<'a>(
 
     Ok(())
 }
+
+/// Mapping of [`PackageRef`]s to the respective index inside the dependency graph.
+pub type LocalPackageIndex = HashMap<PackageRef, (NodeIndex, PathBuf)>;
```
vados-cosmonic, line 847: Can we avoid exposing `petgraph` types outside here? Same comment as that other instance -- if we could wrap the index in some non_exhaustive enum that would probably seal off source of breaking changes and give us some flexibility long term

crates/wasm-pkg-core/src/wit.rs
```diff
@@ -164,7 +174,55 @@ pub fn get_packages(
...
+    let mut graph = DependencyGraph::new();
+    let mut indices = HashMap::new();
+    // establish all nodes
+    for ((spec, path), _) in &pkg_trees {
+        let id = graph.add_node(spec.clone());
+        if indices
+            .insert(spec.package(), (id, path.as_ref().to_owned()))
+            .is_some()
+        {
+            anyhow::bail!("duplicate references to package detected: {spec}");
```
vados-cosmonic, line 211: `anyhow::ensure` feels like a good fit here semantically

crates/wasm-pkg-core/src/wit.rs
```diff
@@ -164,7 +174,55 @@ pub fn get_packages(
...
+        }
+    }
+    for ((spec, _), deps) in pkg_trees {
+        // TODO handle version matching for dependencies
+        for (dep, _version) in deps {
+            if let Some(&(dep, _)) = indices.get(&dep) {
+                let (pkg, _) = indices[&spec.package];
```
vados-cosmonic: lets use `.get` here to avoid panic? Are we somehow guaranteed `&spec.package` is present? a quick comment would be OK instead

crates/wkg/src/main.rs
```diff
@@ -252,37 +257,128 @@ struct PublishArgs {
...
-                package,
-                version: Some(v),
-            }) => Some((package, v)),
-            Some(PackageSpec { version: None, .. }) => {
-                anyhow::bail!("version is required when manually overriding the package ID");
+                let client = self.common.get_client().await?;
+                // 2. Publish our packages in "waves" to the actual registries ensuring all
+                //    possible dependency free pacakges are published in the same group
+                while !plan.is_empty() {
+                    let ready_for_publish = plan.take_ready();
```
vados-cosmonic, line 345: Are we sure this will always terminate? Should we add a check just in case? We're not allowed to have cycles AFAIK so maybe I'm being a bit too defensive here.

crates/wkg/src/main.rs
```diff
@@ -252,37 +257,128 @@ struct PublishArgs {
...
+                                registry: Some(local_registry.clone()),
+                                // we want to publish to "tmp_local_publish" regardless of flags passed in
+                                dry_run: false,
+                            },
+                        )
+                        .await?;
+
+                    let id = plan
+                        .get_node_index(&spec.package)
+                        .expect("missing node index");
```
vados-cosmonic, line 332: Can we use `anyhow::Context` or something here? This is a panic if it somehow goes wrong

crates/wkg/src/main.rs
```diff
@@ -252,37 +257,128 @@ struct PublishArgs {
...
             if lock_file != prev_lock_ref && !self.dry_run {
-                return Err(anyhow::anyhow!(
+                return Err(anyhow!(
                     "wkg.lock would be updated during publish, aborting"
                 ))
                 .with_context(|| {
                     format!(
                         "Run `wkg wit build {}` before attempting to publish",
-                        self.path.to_string_lossy()
+                        path.to_string_lossy()
```
vados-cosmonic:
```suggestion
                        path.display()
```

crates/wkg/src/main.rs
```diff
@@ -312,6 +401,27 @@ impl PublishArgs {
         }
         Ok(())
     }
+
+    fn opts(&self) -> anyhow::Result<PublishOpts> {
```
vados-cosmonic:
```suggestion
    fn publish_opts(&self) -> anyhow::Result<PublishOpts> {
```

crates/wkg/tests/e2e.rs
```diff
@@ -85,6 +89,69 @@ async fn build_and_publish_with_metadata() {
...
+    let namespaces = ["example-a", "example-b", "example-c", "example-d"];
+
+    // copy the transitive-local fixtures from wasm-pkg-core into a temp dir
+    let temp_dir = tempfile::tempdir().expect("Failed to create tempdir");
+    let src_root = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("..")
+        .join("wasm-pkg-core")
+        .join("tests")
+        .join("fixtures")
+        .join("transitive-local");
```
vados-cosmonic: NIT: I think you can do `.join(".../wasm-pkg-core/tests/fixtures/ransitive-local")` ? Am I wrong? that string should turn into a relative path just as well

vados-cosmonic (commented):
Hey I took a pass through, a lot of NITs but it looks mostly good to me!

One thing I'd love to have is a few more tests... This is a pretty substantial feature, so figured there would be more edges/fixtures. That said, we can certainly add tests down the line if we run into anything.

crates/wasm-pkg-common/src/registry.rs
mkatychev: Good point, we can newtype it. Part of the challenge here was the interconnection of wkg crates: cli, client, core, and common. I think I might have exposed more/less than needed. A lot of the visibility changes was a process of fitting.

crates/wasm-pkg-common/src/package.rs
mkatychev: Fair point, I added some of these thinking they were going to be used more often (such as in core & cli) than not 😅.

crates/wasm-pkg-client/src/local.rs
mkatychev: I think this came about because I was double logging with the `"tmp_local_publish"` backend a bit.

Happy to revert. I was thinking through ways of logging selectively, and to your point, log level is not the right place.

crates/wasm-pkg-client/src/local.rs
mkatychev: I feel I trust the end user enough to understand enough of what the `TempDir` struct is in the same way one would understand why holding onto a mutex guard is important. 

If we add it to the struct then it adds surface area overhead for something that feels bespoke to me, is that tradeoff worth it? 🤔

crates/wasm-pkg-client/src/lib.rs
mkatychev: Yes, I agree, a thing we'd need to discuss how to approach. This was more of me reorganizing the logic since I did not grasp the breadth of the change until much later. (I started most of the work with cleaning up `resolve_source`.

mkatychev:
> Hey I took a pass through, a lot of NITs but it looks mostly good to me!
> 
> One thing I'd love to have is a few more tests... This is a pretty substantial feature, so figured there would be more edges/fixtures. That said, we can certainly add tests down the line if we run into anything.

Any particular type of test you had in mind? I think a lot of this will change and I'm leery of being too granular on the testing. If we can add more e2e/smoke scenarios, that would be the best long term approach (less work to be discarded).

crates/wasm-pkg-common/src/package.rs
vados-cosmonic: Yeah so I was looking to see if we actually exposed these to the user... 

To be fair I think we *should* have getters and maybe mark `PackageSpec` as non_exhaustive or something to give ourselves flexibility, but the API is what it is already... so this isn't even the place to discuss that, of course!

crates/wkg/src/main.rs
mkatychev: Perhaps, I actually added that initially though removed afterwards. My rationale that if an error occurred here then the error would be logical rather than technical since `PublishPlan` index layout is set at `from_paths`.

crates/wkg/tests/e2e.rs
mkatychev: Yes, I did that initially though was worried about how windows would handle it (or if the fixture dir structure will change), happy to apply.

crates/wasm-pkg-client/src/local.rs
vados-cosmonic: AH Ok thanks for the heads up -- so I didn't understand that. If we are already logging this message at `info` somewhere else, then IMO you can delete one. Abstractly if I were to guess the "outermost" place should have the "publishing" message... 

Or maybe this is more about alerting on the path that got resolved, rather than publishing, per-say? In that case definitely debug!

crates/wkg/src/main.rs
mkatychev: You're totally right, cargo's `take_ready` bakes in a timeout.
EDIT: forgot to post link:
https://github.com/rust-lang/cargo/blob/a595d0da21f228b7fdae64d3d5c0e527ea66bb59/src/cargo/ops/registry/publish.rs#L289-L294

crates/wasm-pkg-client/src/local.rs
vados-cosmonic: > I feel I trust the end user enough to understand enough of what the TempDir struct is in the same way one would understand why holding onto a mutex guard is important.

I think the idea that the object would be moving around with an invalid path is worth it to remove. The whole "make invalid states unrepresentable" thing -- if you have a `X`, then it is valid/unlikely to have been silently corrupted somehow.

We can make the `TempDir` itself not `pub` so hat the user doesn't know it's there? We can't stop them from cloning the path or doing stuff based on it of course.

crates/wkg/src/main.rs
vados-cosmonic: Oh Ok great! but it's not like... async? I think I actually have more questions now -- I was wondering if it's possible for `ready_for_publish` to not have anything in it and the plan to *not* be empty, essentially.

Not looking for a proof or anything of course, but basically... could the graph be misconfigured in a way that nodes are kinda hanging out but don't end up looking ready? The answer could easily be "absolutely not"  and then this isn't an issue, but a small invariant check (i.e. "if plan is empty, take_ready will always produce at least one thing to do on each iteration") would be nice

crates/wasm-pkg-client/src/local.rs
mkatychev: Good point, will do.

vados-cosmonic:
Yeah so it gets kinda sticky real quick but what a about a test for nested folders? I *think* the test right now all the published packages are at the same level, IIRC you can define new packages in nested folders, I'm wondering if those would get published too.

mkatychev:
> IIRC you can define new packages in nested folders, I'm wondering if those would get published too.

They have in the tests I did. I guess I have to upload those test cases too.

Good suggestion!

crates/wkg/src/main.rs
```diff
@@ -252,37 +257,128 @@ struct PublishArgs {
...
+            [path] => path,
+            paths => {
+                let mut overlay_config = self.common.load_config().await?;
+                let cache = self.common.load_cache().await?;
+
+                let (local_config, _tmp_dir_handle) = LocalConfig::temp_dir()?;
+                let reg_config =
+                    RegistryConfig::default().with_default_backend(LOCAL_PROTOCOL, local_config)?;
+                // Route every package in the plan to the local overlay registry
+                // backed by `reg_config`, so the client used in `build_wit_dir`
```
mkatychev, line 271: `reg_config` no longer exists, remove

crates/wasm-pkg-client/src/local.rs
mkatychev: Oh, I recall the problem here, `LocalConfig` is stored in serialized form and the `<T: Serlialize> impl Drop for T` would run on `set_backend_config` close.

mkatychev: Part of the challenge is that the backend is instantiated every time for `resolve_source` meaning that we don't have an opportunity to hold onto the handles:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/ce353f6d02339f8b848a1c4667099581a6f2e93b/crates/wasm-pkg-client/src/lib.rs#L288-L288

crates/wasm-pkg-client/src/local.rs
mkatychev: I think it should stay `info` for now and we can deal with the noise later, the point is that with a local backend multi package package flow for a registry called `foo`:
```toml
default_registry = "foo"

[registry.foo.local]
root = "./foo_dir"
```

we would end up calling that line twice per package, once for `"tmp_local_publish"`:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/ce353f6d02339f8b848a1c4667099581a6f2e93b/crates/wkg/src/main.rs#L319-L319

... and one on the real `LocalBackend` of `"foo"`:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/ce353f6d02339f8b848a1c4667099581a6f2e93b/crates/wkg/src/main.rs#L346-L349

crates/wasm-pkg-common/src/package.rs
mkatychev: so the verdict is removing that method for now?

crates/wasm-pkg-common/src/registry.rs
mkatychev: I moved it to common as a `pub(crate)` type, that seems to have taken care of the exposing.

crates/wasm-pkg-common/src/package.rs
vados-cosmonic: IMO yeah, as users already had access to the struct members, I think they can call clone themselves!

crates/wkg/src/main.rs
mkatychev: Here's what I've reasoned myself into:
* `ready_for_publish` is guaranteed to be nonempty IF `!plan.is_empty()` and 
  * a `DependencyGraph` (`petgraph::Acyclic`) will produce `AcyclicEdgeError` errors during `try_update_edge` for any "danging graphs"[^1]
* a node is determined to be ready when it has zero `Incoming` edges[^2]:
  https://github.com/bytecodealliance/wasm-pkg-tools/blob/c26b6bfdc351769427b21c12e06a9fc72779acdc/crates/wasm-pkg-core/src/resolver.rs#L899-L902
  
[^1]: https://github.com/bytecodealliance/wasm-pkg-tools/blob/8555d1bbee9ac20290b5516437f35346add95be0/crates/wasm-pkg-core/src/wit.rs#L216-L227
[^2]: https://docs.rs/petgraph/latest/src/petgraph/graph_impl/mod.rs.html#913


The one area we _don't_ have guarantees is if `PackagePublisher::publish` will ever terminate.

crates/wkg/src/main.rs
```diff
@@ -252,37 +257,128 @@ struct PublishArgs {
...
+                        let id = plan
+                            .get_node_index(&spec.package)
+                            .expect("missing node index");
+                        let source = Box::pin(Cursor::new(validated_packages[&id].clone()));
+                        let (package, version) = client
+                            .client()?
+                            .publish_release_data(source, publish_opts.clone())
+                            .await?;
+                        if self.dry_run {
+                            println!("Aborting publish due to dry run: {}@{}", package, version);
```
vados-cosmonic:
```suggestion
                            eprintln!("Aborting publish due to dry run: {}@{}", package, version);
```

This message and other diagnostic messages should be printed to STDERR

crates/wasm-pkg-core/src/wit.rs
```diff
@@ -164,7 +183,62 @@ pub fn get_packages(
...
+    Ok((package, packages))
+}
+
+/// Build an acyclic [`DependencyGraph`] for the provided WIT package paths alongside a [`LocalPackageIndex`].
+///
+/// # Errors
+///
+/// The function will return an error if there are duplicate references to the same package or if
+/// there are cyclical dependencies between packages.
+pub fn get_local_dependencies(
```
vados-cosmonic: I think this can be `pub(crate)` -- I'm only seeing it be used in `wasm-pkg-core`?

crates/wasm-pkg-core/src/resolver.rs
vados-cosmonic: Hey @mkatychev I think GH hid these comments from you -- did you happen to get this one?

crates/wkg/src/main.rs
mkatychev: Good call

mkatychev:
For https://github.com/bytecodealliance/wasm-pkg-tools/pull/215#discussion_r3478632273

Perhaps I resolved a comment that I didn't actually address, if so then my apologies.

mkatychev:
@vados-cosmonic for `PublishPlan`, the need for `pub` is because it is instantiated inside of wkg proper, this is also true for all the methods within and inner types (except for `PublishPlan::len` but that one makes sense to be exposed):
https://github.com/bytecodealliance/wasm-pkg-tools/blob/778b1cc6ba025ec2b9f4b976b760029edc9d8213/crates/wkg/src/main.rs#L279

crates/wkg/src/main.rs
mkatychev: @vados-cosmonic does the explanation make sense or should we move it back to emittiong an error?

vados-cosmonic: Yeah that makes sense to me!

crates/wasm-pkg-core/src/resolver.rs
vados-cosmonic: Hey @mkatychev this one is still relevant -- we expose `LocalPackageIndex` via `PublishPlan`, and it contains `NodeIndex` (available via `get_node_index`) which is from `petgraph` which will tie us to their breaking changes. Not the end of the world, but would be nice to avoid it.

crates/wasm-pkg-core/src/wit.rs
vados-cosmonic: This isn't that big of a deal, let's skip it

crates/wkg/src/main.rs
vados-cosmonic: IMO @mkatychev this would make for a great comment close in the code!

crates/wasm-pkg-core/src/resolver.rs
vados-cosmonic: I think we can change this in the future if need be -- we're not at a 1.x version release just yet.

vados-cosmonic (approved):

PR #219, opened 2026-06-30
refactor(cli): move `wkg wit <cmd>` subcommands to `wkg <cmd>`

mkatychev:
* add depracation warning to existing `wkg wit` commands

Current spead of subcommands for `wkg` is unintuitive with two top level commands that only work on WIT:

* `wkg publish`
* `wkg get`

The `wkg wit` namespace scoping no longer makes sense as Warg is being phased out in favour of OCI.

### DIFF
--- a/crates/wkg/src/wit.rs
+++ b/crates/wkg/src/wit.rs
@@ -16,19 +16,8 @@ use crate::Common;
 /// Commands for interacting with wit
 #[derive(Debug, Subcommand)]
 pub enum WitCommands {
-    /// Build a WIT package from a directory. By default, this will fetch all dependencies needed
-    /// and encode them in the WIT package. This will generate a lock file that can be used to fetch
-    /// the dependencies in the future.
     Build(BuildArgs),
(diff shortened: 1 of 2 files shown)


crates/wkg/src/wit.rs
```diff
@@ -57,6 +49,12 @@ pub struct BuildArgs {
     pub common: Common,
 }
 
+/// Fetch dependencies for a component. This will read the package containing the world(s) you
```
vados-cosmonic, line 52: Do these show up in usage output? I'm not sure if the comments around `Args` deriving structs are used the same way as comments in `Subcommand` deriving structs

mkatychev: They sure do! I tested just to make sure. This way we don't duplicate comments. Isn't clap neat?

vados-cosmonic: Thanks for testing! That's a great feature -- glad clap had us covered here.

vados-cosmonic (approved):

PR #220, opened 2026-06-30
feat(cli): Added `--color` styling flag to wkg cli

mkatychev:
Added `--color` styling flag to wkg cli:
```sh-session
$ ./target/debug/wkg --help
Usage: wkg [OPTIONS] <COMMAND>

Commands:
  ...
Options:
      --color <WHEN>  Controls when to use color [default: auto] [possible values: auto, always, never]
  ...

```

Added `warnln!`, `statusln!`, and `helpln!` (stolen™ from cargo):
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,20 +9,31 @@ authors = ["The Wasmtime Project Developers"]
 license = "Apache-2.0 WITH LLVM-exception"
 
 [workspace.dependencies]
+wasm-pkg-client = { version = "0.16.0", path = "crates/wasm-pkg-client" }
+wasm-pkg-common = { version = "0.16.0", path = "crates/wasm-pkg-common" }
+wasm-pkg-core = { version = "0.16.0", path = "crates/wasm-pkg-core" }
+
+anstream = "0.6"
+anstyle = "1.0"
 anyhow = "1"
+async-trait = "0.1.77"
 base64 = "0.22"
 bytes = "1.11"
+clap = "4.5"
+colorchoice-clap = "1.0"
 docker_credential = "1.3.2"
 etcetera = "0.11"
 futures-util = "0.3.30"
-oci-client = { version = "0.16", default-features = false, features = [
-    "rustls-tls",
-] }
-oci-wasm = { version = "0.4", default-features = false, features = [
-    "rustls-tls",
-] }
+glob = "0.3.3"
+http = "1.1.0"
+indexmap = "2.5"
+oci-client = { version = "0.16", default-features = false, features = [ "rustls-tls" ] }
+oci-wasm = { version = "0.4", default-features = false, features = [ "rustls-tls" ] }
 petgraph = "0.8.3"
 rcgen = "0.14.8"
+reqwest = { version = "0.12.0", default-features = false }
+rstest = "0.23"
+secrecy = "0.8"
 semver = "1.0.23"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1"
@@ -34,31 +45,14 @@ tokio = "1.44.2"
 tokio-util = "0.7.10"
 toml = "0.8"
 tracing = "0.1.40"
-tracing-subscriber = { version = "0.3.20", default-features = false, features = [
-    "fmt",
-    "env-filter",
-] }
-wasm-pkg-common = { version = "0.16.0", path = "crates/wasm-pkg-common" }
-wasm-pkg-client = { version = "0.16.0", path = "crates/wasm-pkg-client" }
-wasm-pkg-core = { version = "0.16.0", path = "crates/wasm-pkg-core" }
-wasm-metadata = "0.244"
-wit-component = "0.244"
-wit-parser = "0.244"
-anstream = "0.6"
-anstyle = "1.0"
-async-trait = "0.1.77"
-clap = "4.5"
-colorchoice-clap = "1.0"
-glob = "0.3.3"
-http = "1.1.0"
-indexmap = "2.5"
-reqwest = { version = "0.12.0", default-features = false }
-rstest = "0.23"
-secrecy = "0.8"
+tracing-subscriber = { version = "0.3.20", default-features = false, features = [ "fmt", "env-filter" ] }
 url = "2.5.0"
 warg-client = "0.9.2"
 warg-crypto = "0.9.2"
 warg-protocol = "0.9.2"
+wasm-metadata = "0.244"
+wit-component = "0.244"
+wit-parser = "0.244"
 
 # https://github.com/crate-ci/typos/blob/master/docs/reference.md
 [workspace.metadata.typos.default]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,9 +1,9 @@
 [workspace]
 members = ["crates/*"]
-resolver = "2"
+resolver = "3"
 
 [workspace.package]
-edition = "2021"
+edition = "2024"
 version = "0.16.0"
 authors = ["The Wasmtime Project Developers"]
 license = "Apache-2.0 WITH LLVM-exception"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,21 @@ wasm-pkg-core = { version = "0.16.0", path = "crates/wasm-pkg-core" }
 wasm-metadata = "0.244"
 wit-component = "0.244"
 wit-parser = "0.244"
+anstream = "0.6"
+anstyle = "1.0"
+async-trait = "0.1.77"
+clap = "4.5"
+colorchoice-clap = "1.0"
+glob = "0.3.3"
+http = "1.1.0"
+indexmap = "2.5"
+reqwest = { version = "0.12.0", default-features = false }
+rstest = "0.23"
+secrecy = "0.8"
+url = "2.5.0"
+warg-client = "0.9.2"
+warg-crypto = "0.9.2"
+warg-protocol = "0.9.2"
 
 # https://github.com/crate-ci/typos/blob/master/docs/reference.md
 [workspace.metadata.typos.default]
--- a/crates/wasm-pkg-common/src/package.rs
+++ b/crates/wasm-pkg-common/src/package.rs
@@ -7,14 +7,14 @@ use crate::{label::Label, Error};
 pub use semver::Version;
 
 #[cfg(feature = "anstyle")]
-mod ansi {
+pub(crate) mod ansi {
     use anstyle::{Ansi256Color, AnsiColor, Style};
 
-    pub const LABEL: Style = AnsiColor::BrightBlue.on_default().bold();
-    pub const VERSION: Style = AnsiColor::BrightRed.on_default();
-    pub const SEP: Style = Ansi256Color(249).on_default();
+    pub(crate) const LABEL: Style = AnsiColor::BrightBlue.on_default().bold();
+    pub(crate) const VERSION: Style = AnsiColor::BrightRed.on_default();
+    pub(crate) const SEP: Style = Ansi256Color(249).on_default();
 
-    pub fn is_terminal() -> bool {
+    pub(crate) fn is_terminal() -> bool {
         use std::io::IsTerminal;
         std::io::stderr().is_terminal()
     }
--- a/crates/wasm-pkg-common/src/package.rs
+++ b/crates/wasm-pkg-common/src/package.rs
@@ -6,7 +6,7 @@ use crate::{label::Label, Error};
 
 pub use semver::Version;
 
-#[cfg(feature = "anstyle")]
+#[cfg(feature = "ansi-term-output")]
 pub(crate) mod ansi {
     use anstyle::{Ansi256Color, AnsiColor, Style};
 
@@ -49,7 +49,7 @@ impl PackageRef {
 
 impl std::fmt::Display for PackageRef {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        #[cfg(feature = "anstyle")]
+        #[cfg(feature = "ansi-term-output")]
         if ansi::is_terminal() {
             use ansi::{LABEL, SEP};
             return write!(
@@ -93,7 +93,7 @@ impl FromStr for PackageRef {
 }
 impl std::fmt::Display for PackageSpec {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        #[cfg(feature = "anstyle")]
+        #[cfg(feature = "ansi-term-output")]
         if ansi::is_terminal() {
             use ansi::{SEP, VERSION};
             return match &self.version {
--- a/crates/wasm-pkg-common/src/package.rs
+++ b/crates/wasm-pkg-common/src/package.rs
@@ -2,11 +2,11 @@ use std::str::FromStr;
 
 use serde::{Deserialize, Serialize};
 
-use crate::{label::Label, Error};
+use crate::{Error, label::Label};
 
 pub use semver::Version;
 
-#[cfg(feature = "ansi-term-output")]
+#[cfg(all(feature = "ansi-term-output", not(feature = "test")))]
 pub(crate) mod ansi {
     use anstyle::{Ansi256Color, AnsiColor, Style};
 
@@ -44,7 +44,7 @@ impl PackageRef {
 
 impl std::fmt::Display for PackageRef {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        #[cfg(feature = "ansi-term-output")]
+        #[cfg(all(feature = "ansi-term-output", not(feature = "test")))]
         {
             use ansi::{LABEL, SEP};
             write!(
@@ -53,7 +53,7 @@ impl std::fmt::Display for PackageRef {
                 self.namespace, self.name,
             )
         }
-        #[cfg(not(feature = "ansi-term-output"))]
+        #[cfg(not(all(feature = "ansi-term-output", not(feature = "test"))))]
         write!(f, "{}:{}", self.namespace, self.name)
     }
 }
@@ -91,7 +91,7 @@ impl std::fmt::Display for PackageSpec {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self.version {
             Some(version) => {
-                #[cfg(feature = "ansi-term-output")]
+                #[cfg(all(feature = "ansi-term-output", not(feature = "test")))]
                 {
                     use ansi::{SEP, VERSION};
                     write!(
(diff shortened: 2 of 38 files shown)


mkatychev:
Depends on https://github.com/bytecodealliance/wasm-pkg-tools/pull/219

vados-cosmonic (commented):
I'm not against using `anstyle` for this, but what do you think about using `tracing` w/ `tracing_subscriber::fmt`'s ANSI color layer? That would give us error/warn/info colors, and be more interoperable with tracing in the libs.

mkatychev:
> I'm not against using `anstyle` for this, but what do you think about using `tracing` w/ `tracing_subscriber::fmt`'s ANSI color layer? That would give us error/warn/info colors, and be more interoperable with tracing in the libs.

I've thought about that as well, when we get to exposing wkg as a lib, I would rather have switch to tracing as a detection layer similar to how cargo's [`Shell::print`] does it:
https://github.com/rust-lang/cargo/blob/a595d0da21f228b7fdae64d3d5c0e527ea66bb59/crates/cargo-util-terminal/src/shell.rs#L79-L96

that flexibility needs to be built to and should come with a [`GlobalContext` analogue](https://github.com/rust-lang/cargo/blob/a595d0da21f228b7fdae64d3d5c0e527ea66bb59/src/cargo/util/context/mod.rs#L211-L213) as well (which we need for better workspace support as well).

I think for now it makes sense to distinguish between purely human stuff (no tracing wrapper) and things triggered by `RUST_LOG`, ultimately all the `anstyle` stuff should either be turned off or turned into `tracing::event!` calls when used as a lib.

crates/wasm-pkg-common/src/package.rs
```diff
@@ -70,10 +83,15 @@ impl FromStr for PackageRef {
...
 impl std::fmt::Display for PackageSpec {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        if let Some(version) = &self.version {
-            write!(f, "{}@{}", self.package, version)
-        } else {
-            write!(f, "{}", self.package)
+        match (&self.version, f.alternate()) {
+            (Some(version), true) => write!(
+                f,
+                "{:#}{SEP}@{SEP:#}{VERSION}{version}{VERSION:#}",
```
vados-cosmonic: This seems like it might surprise people who are using `PackageSpec` from here as a library -- I don't normally expect alternate display impls to write terminal escape codes.

Does it make sense to put this behind a feature flag?

crates/wasm-pkg-common/src/package.rs
```diff
@@ -1,11 +1,16 @@
 use std::str::FromStr;
 
+use anstyle::{Ansi256Color, AnsiColor, Style};
 use serde::{Deserialize, Serialize};
 
 use crate::{label::Label, Error};
 
 pub use semver::Version;
 
+const LABEL: Style = AnsiColor::BrightBlue.on_default().bold();
```
vados-cosmonic: Can we put this behind a feature flag? IMO `anstyle` should be behind one too for this lib

crates/wasm-pkg-core/src/resolver.rs
```diff
@@ -955,8 +955,13 @@ impl std::fmt::Display for PublishPlan {
                 .peekable();
 
             if neighbors.peek().is_none() {
-                // tracing::debug!("{dep} has no dependents");
-                writeln!(f, "[{dep} has no dependents]")?;
+                if alt {
```
vados-cosmonic: Same note here about feature flagging color output code

crates/wkg/src/main.rs
vados-cosmonic, line 1: can we check for whether we're [in a terminal](https://doc.rust-lang.org/beta/std/io/trait.IsTerminal.html#tymethod.is_terminal) before using the escape codes? Does `anstyle` automatically handle this?

crates/wasm-pkg-common/src/package.rs
mkatychev: Good point!

crates/wkg/src/main.rs
mkatychev: I can look into it

crates/wasm-pkg-common/src/package.rs
mkatychev: Should `.alternate()` do _anything_ here if say a colour/ANSI flag feature  gates this?

vados-cosmonic: Probably not IMO -- the impl could look like the previous one here, I don't know if we have much to expand/more clearly print. 

I *usually* use that alternative mechanism for getting more debug level display/more involved display and I don't think there's that much to even do here.

mkatychev: Agreed, to your point of keeping this out of libs, is there any way we can use `anstyle` in the bin and keep it out of the lib whilst still getting this granular control? 🤔

vados-cosmonic: A feature flag would work, no? The lib can pull in the version with the feature flag (triggering some version of `anstyle` dep in the lib), and get colored output. 

Alternatively, we can have the bin use a wrapper type and implement it's own `Display` and use *that*, cleaner, but harder to integrate because it's likely going to cascade to more types needing to be wrapped. 

Maybe we just don't need to color this output?

crates/wkg/src/main.rs
```diff
@@ -32,9 +33,40 @@ use wit::{BuildArgs, FetchArgs, UpdateArgs, WitCommands};
...
+        let style = ::anstyle::AnsiColor::Magenta.on_default() | ::anstyle::Effects::BOLD;
+        ::anstream::eprintln!("{style}{:>7}{style:#}: {}", "help", format_args!($($arg)*));
+    }};
+}
+
 #[derive(Parser, Debug)]
 #[command(version)]
 struct Cli {
+    #[command(flatten)]
+    color: colorchoice_clap::Color,
```
mkatychev, line 68: I thought about feature gating this behind an ansi/colour flag but that currently does not make sense since `tracing` already produces ansi/colour styling

crates/wasm-pkg-core/Cargo.toml
```diff
@@ -7,10 +7,13 @@ version.workspace = true
 authors.workspace = true
 license.workspace = true
 
+[features]
+anstyle = ["wasm-pkg-common/anstyle"]
```
vados-cosmonic: Would you mind naming this something more meaningful? `anstyle` is the crate but I don't think it's obvious what this controls. Compare this to naming it something like `colored-output`, which *may* pull in anstyle as a feature off of that (I doubt that we ever need to have stuff like `anstyle_0_7`, etc, but that's IMO a good use of a crate-named feature).

crates/wkg/src/main.rs
vados-cosmonic: Hey yeah so we don't need color flags IMO for the binary -- that makes sense as being implicitly expected. IMO the only thing we need to do is check for terminal properly when we're writing output, etc.

crates/wkg/src/main.rs
```diff
@@ -32,9 +33,40 @@ use wit::{BuildArgs, FetchArgs, UpdateArgs, WitCommands};
...
+        let style = ::anstyle::AnsiColor::Yellow.on_default() | ::anstyle::Effects::BOLD;
+        ::anstream::eprintln!("{style}warning{style:#}: {}", format_args!($($arg)*));
+    }};
+}
+
+#[macro_export]
+macro_rules! statusln {
+    ($label:expr, $($arg:tt)*) => {{
+        let style = ::anstyle::AnsiColor::Cyan.on_default() | ::anstyle::Effects::BOLD;
+        ::anstream::eprintln!(
```
vados-cosmonic, line 48: Can we check for terminal-hood here?

vados-cosmonic (commented):
Thanks for taking another pass at this, we're *almost there*.

crates/wasm-pkg-core/Cargo.toml
mkatychev: one concern here is that we have the possiblity (based on upstream configs) of emitting something that is bolded but colourless, maybe `ansi-output`?

crates/wkg/src/main.rs
mkatychev: and call `println!` if no terminal or nothing at all?

EDIT: or a `tracing::event!` 🤔 but I wonder if decision makes more sense at the feature-flag level...

crates/wasm-pkg-core/Cargo.toml
vados-cosmonic: Yeah so I'd say in that case then we might call it something like `pretty` or `pretty-output`?

`ansi-output` is good but IMO it doesn't really impart that you're going to add bolding/color/use *for the purpose of making the output more legible*. I mean there's a lot to say to try to get that idea across and we only have so much space... That said, I'm not super tied to what it's called, as long as it gets the intent across, not the how.

crates/wkg/src/main.rs
vados-cosmonic: yup! `eprintln` in this case, but yeah that sounds good -- You might even be able to just set the style to empty string? `is_terminal().then_some(|| ansicolor... ).unwrap_or_default()`? something like that.

IMO we can hold off on tracing until some other time -- I think your ideas about the separation/how to do it when we talked were good, but those changes should probably come in like a big block!

mkatychev: agreed!

vados-cosmonic: yeah a regular `if` is probably cleaner... Just wish we could  do it once and then never deal with it again somehow rather than introducing a branch, but I think that's just the nature of things

mkatychev: Actually, `anstream` already seems to be handing `is_terminal` internally:

1. [`AutoStream` handling `is_terminal`](https://github.com/rust-cli/anstyle/blob/d804bdc728525994f45c06c6219ce7ed10f6a770/crates/anstream/src/auto.rs#L173)
1. [`anstream::eprinln!` invoke of `to_adapted_string`](https://docs.rs/anstream/latest/src/anstream/_macros.rs.html#188)
1. [`AutoStream` used in `to_adapted_string`](https://docs.rs/anstream/latest/src/anstream/_macros.rs.html#357)

crates/wasm-pkg-core/Cargo.toml
vados-cosmonic: `pretty-term`? `ansi-term-output`?

crates/wkg/src/main.rs
mkatychev: Just updated my comment above with sequential relevance @vados-cosmonic, we should not need to handle `is_terminal` at all if we use `anstream` macros since they do it for us.

crates/wasm-pkg-core/Cargo.toml
mkatychev: > `pretty-term`? `ansi-term-output`?

I think my holdover is mainly the `-term` section, see https://github.com/bytecodealliance/wasm-pkg-tools/pull/220#discussion_r3510053292. There are special pseudo-tty (heh) uses where ANSI codes get handled without being in a "terminal":
* CI logs: https://github.com/dandavison/delta/issues/1310 (GHA)
* `.log` filetype renderers: 
  * https://plugins.jetbrains.com/plugin/29008-ansi-log-viewer
  * https://typst.app/universe/package/ansi-render/


This feels like splitting hairs but if you feel `ansi-term-output` is a good enough approximation, then let's go with it.

mkatychev: ...maybe let's just go with `ansi-term-output` 🫠

vados-cosmonic: Sounds good! We've definitely painted the shed here :)

crates/wkg/src/main.rs
vados-cosmonic: Thanks for pointing this out! glad that we don't have to do it!

crates/wasm-pkg-core/Cargo.toml
vados-cosmonic: Woops, sorry closed this early, but once this is resolved I'm good to ✅  and you can get this merged!

mkatychev: one last thing, per https://github.com/bytecodealliance/wasm-pkg-tools/pull/220#discussion_r3510053292, I've removed uses of `is_terminal` to keep things simple (and since the feature flag is now called `ansi-term-output`):
https://github.com/bytecodealliance/wasm-pkg-tools/pull/220/commits/cbc3feacc695077a05dee01123a13bb656f0f0fa

EDIT: there are workspace level test failures when using the feature resolver

Cargo.toml
```diff
@@ -1,28 +1,39 @@
 [workspace]
 members = ["crates/*"]
-resolver = "2"
+resolver = "3"
 
 [workspace.package]
-edition = "2021"
+edition = "2024"
```
mkatychev, line 6: Tried seeing if bumping would resolve previous test failures:
https://github.com/bytecodealliance/wasm-pkg-tools/actions/runs/28563549597/job/84686098906

Updating edition did not fix issues and this is out of scope but a nice to have regardless.

This was actual fix:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/f5ded81fe92425e0a2cce4b7f63eafb0559a9973/crates/wasm-pkg-common/Cargo.toml#L18-L19

TLDR: resolver 3 (and 2) bleeds `wkg` feature flags when running `cargo test` from the workspace.

`cargo test -p wasm-pkg-core` is not impacted but it is impractical to invoke explicitly all the time - hence `"test"` feature flag for ergonomics.

mkatychev: alternative: we can remove the ansi styling completely from the types but regaining functionality (if worth it) will not be feasible short term and the workspace feature bleed will still be a problem for other flags.

vados-cosmonic: To try and restate the problem here -- was it that `cargo test -p wkg` (or regular `cargo test` of the workspace) dependency-unifies to `wasm-pkg-core *with* `ansi-term-output` enabled, no matter what, because of the dep w/ features at the `wkg` level? 

If so, IMO I don't think we even need to introduce the `test` feature, adding `cargo test -p wasm-pkg-core --no-default-features` or something to CI to make sure it runs without is enough.

We could even pull in something like [cargo-hack](https://github.com/bytecodealliance/wasm-pkg-tools/pull/220/changes#r3510620863) for CI in the future, but my concern here is more that we just get it covered under test so we can't put out a broken release/merge broken code to main.

It's a style thing but what do you think about defining the dep and it's transitive flag dependencies explicitly in `wkg` too, but as a default?

I believe the syntax would be something like:

```toml
[features]
defaults = ["ansi-term-output"]
ansi-term-output = [ "wasm-pkg-common/ansi-term-output", "wasm-pkg-core/ansi-term-output"]
```

and then we could make the ansi-term output deps optional like the other lib crates (though they are pulled in by default)

If someone for example wanted to build the binary even smaller, for some reason, they could run `cargo install wkg --no-default-features` and get a non-ansi-term build this way, rather than the dependencies being required.

I think this pattern is pretty well accepted, as far as optional deps go, even for bin crates, but I'm OK if you don't want to do this either.

vados-cosmonic (approved):
LGTM 🚀

mkatychev:
> If so, IMO I don't think we even need to introduce the `test` feature, adding `cargo test -p wasm-pkg-core --no-default-features` or something to CI to make sure it runs without is enough.

From a UX perspective, I feel it is important for a default argument set to behave predictably and avoid external dependencies _as much as possible_ (especially for top level subcommands like `* build`, `* check`, and `* test`), it is why this section bothers me so much:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/23cbf8f00838d6689c62e58440057348c367379e/crates/wkg/Cargo.toml#L11-L14

I would rather deal with awkward design choices in the margins than rely on callouts & caveats sprinkled everywhere to push the user towards providing a specific set of parameters or lean on external tooling like docker, cargo-hack, bash, or just to scaffold baseline expectations.

This is obviously subjective but a user should be surprised with a failure when they run something like `cargo test` on a repo after a fresh clone.

crates/wkg/src/oci.rs
```diff
@@ -235,7 +235,10 @@ mod tests {
         // NOTE(thomastaylor312): These have to run serially because we are setting an env var
         into_auth_should_read_docker_registry_credentials();
         into_auth_should_other_registry_credentials();
-        std::env::remove_var("DOCKER_CONFIG");
+        // NOT-SAFE: this is likely causing race conditions
```
mkatychev, line 238: I run into race conditions on darwin arm very often with these tests, @ryan-surname-p have you had the same experience?

vados-cosmonic:
Yeah unfortunately I think how Rust work has our hands tied here, given these facts:

1. Using the `ansi-term-output` feature in wkg  pulls in extra crates
2. `ansi-term-output` isn't actually required, but we want it to be enabled by default

I'd like to have this be really obvious from reading the Cargo.toml.

Separately, the `docker-tests` feature should not exist (IMO, like the `test` feature shouldn't) -- the most elegant way I've seen that navigated is discovering those features in `build.rs` and injecting a dynamic [`cfg`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-cfg) or ENV that lets the tests do what they need to do. I think the current code was an initial approach at this that should be removed. 

As far as what the user (developer) tests/has to configure, there is no actual change here based on what I'm describing, it's purely notational, except for the user losing the ability to build *without* those features required. Requiring those features without giving an opt out is the same as having it as a default feature, except slightly worse because the user has no way to do a more minimal build.

The onus to test both the default build and the minimal build is on us as maintainers IMO -- the average developer need not do it.

mkatychev:
@vados-cosmonic thanks for the explanation on https://github.com/bytecodealliance/wasm-pkg-tools/pull/220#issuecomment-4866424022

crates/wkg/src/oci.rs
ryan-surname-p: I don't recall having issues with this but that's the fun thing about race conditions I could have just gotten lucky. 

That said this is definitely racy which it seems someone has acknowledged above on line 235 as well. 

Wrapping it in unsafe does make it much more clear.

vados-cosmonic:
@mkatychev  Oh no worries, the `build.rs` feature detection thing is it's a pattern I've seen others do and shamlessly stole 😆 

It's funny how every time something just-barely-uncomfortable like this comes up I go search on rust-lang/rust and find some 5+ year old issue where someone is asking for the same thing :)

mkatychev:
I think I fell for the classic scope creep trap in this PR, I will revert the core/common changes and merge without the display impls and submit a separate PR

crates/wkg/src/oci.rs
mkatychev: the `unsafe` was a hard error thanks to the 2024 edition bump!

PR #222, opened 2026-07-02
ci: add caching for rust jobs

mkatychev:
Added `rust-toolchain.toml` as well

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,14 +23,14 @@ jobs:
           - os: macos-latest
             docker_tests_flags: "--no-default-features"
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       # selecting a toolchain either by action or manual `rustup` happen
       # before cache setup
       - run: rustup toolchain install stable --profile minimal
 
       - name: Set up rust cache
-        uses: Swatinem/rust-cache@v2
+        uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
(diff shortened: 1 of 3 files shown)


rust-toolchain.toml
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "stable"
+targets = ["wasm32-wasip1"]
```
mkatychev, line 3: Should we bump this to p2?

.github/workflows/ci.yml
```diff
@@ -23,10 +23,17 @@ jobs:
           - os: macos-latest
             docker_tests_flags: "--no-default-features"
     steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: actions/checkout@v7
```
vados-cosmonic: let's pin to sha hashes along with the versions in comments

.github/workflows/publish-binaries.yml
mkatychev, line 1: wonder if publish-binaries should be done in this PR as well 🤔

rust-toolchain.toml
vados-cosmonic: Can we add p2 alongside? IIRC when building P3 components we still build p1 first then adapt our way there, due to upstream toolchain mismatches

vados-cosmonic (commented):
Thanks!

mkatychev:
https://github.com/bytecodealliance/wasm-pkg-tools/pull/222#discussion_r3515255660

Unless we explicitly call both p2 and p1 as targets, we should not gain any benefits here, rustup will just spend more time downloading. Perhaps orthogonalize the tests?

vados-cosmonic:
Yeah in this case then let's just leave it P1 and consider later!

.github/workflows/publish-binaries.yml
vados-cosmonic: Let's push that to another PR!

Issue #171, opened 2025-08-21
Add support for signing and validating OCI packages


lann:
https://docs.sigstore.dev/cosign/signing/signing_with_containers/

WASI packages are already being signed as part of the publish workflow: https://github.com/WebAssembly/wasi-http/blob/ed620562ee1d089a20aa67cd7790782d03f2f505/.github/workflows/publish-0.3.yml#L71-L72

It should be relatively straightforward to validate these signatures with the [`sigstore` crate](https://docs.rs/sigstore/latest/sigstore/index.html).

ryan-surname-p:
I have an active branch which can perform [sigstore bundle](https://docs.sigstore.dev/about/bundle/) verification (based on the current [wasi package signing](https://github.com/WebAssembly/WASI/blob/v0.2.12/.github/workflows/publish-proposal.yml#L76-L82) policy).

Currently I have wkg perform verification as a stand alone command off the fetch path but I’m leaning toward leaving that as a job for `cosign`.

After implementing it though I am not quite sure i's a great idea to pull directly into wkg.

The "[official](https://docs.sigstore.dev/language_clients/rust/)" rust client is experimental and does not support attestation verification.

 The sigstore [crate](https://github.com/sigstore/sigstore-rust) which does support attestations verification is mostly managed by a single individual. 

IMO it's probably best to just keep the act of package signing and verification out of wkg as it could quickly become a maintenance burden which detracts from core wkg development.

mkatychev:
I think it's fine to include it as an OCI metadata key, that way we don't have to explicitly support crypto logic, right now metadata has [explicitly used fields](https://tag-runtime.cncf.io/wgs/wasm/deliverables/wasm-oci-artifact/) (like `oci.artifact.type: application/vnd.wasm.config.v0+json`)

ryan-surname-p:
For sure, I think supporting the [Referrers API ](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers) could have quite a few benefits outside of just signatures.

With oci-client 0.17.0 the referrers fallback logic is handled automatically for you which was required to discover the signature bundles on ghcr.io

mkatychev:
Thanks for the reference 🥁!

PR #225, opened 2026-07-06
bump oci-client 0.16→0.17 and oci-wasm 0.4→0.5

ryan-surname-p:
Popular registries, namely ghcr, do not fully implement the OCI 1.1 Referrers API. 

Version 0.17 implements automatic tag-schema fallback as described [here](https://opencontainers.org/posts/blog/2024-03-13-image-and-distribution-1-1/#support-for-existing-registries).

This is necessary when we want to validate DSSE bundles such as the ones bound to `wasi/http:0.2.12`  (tangential to #171)


ryan-surname-p:
As an aside I wonder if it would be worth re-exporting the oci-client bits wkg cares about through wasm-oci so you only need to worry about the one dependency.

mkatychev (approved):
LGTM

PR #228, opened 2026-07-10
chore(deps): update wasm tools

vados-cosmonic:
Upstream wasm tools has come quite a ways since it was last updated, so this PR updates it. 

There are some places where the interface has legitimately changed, but it seems like the biggest difference is more specificity around errors.


mkatychev (approved):
LGTM

PR #229, opened 2026-07-10
clarify single-world requirement in extract_resolve_and_world_id

ryan-surname-p:
The original wording was factually inaccurate as wit_parser will not arbitrarily select a world from a wit package. 

Furthermore semver checks are performed on a per component basis so the wit package being decoded must too define as a single world. 

https://component-model.bytecodealliance.org/design/wit.html#worlds

>  a world describes the contract of a component. A world describes a set of imports and exports…

### DIFF
--- a/crates/wasm-pkg-client/src/decoded_component.rs
+++ b/crates/wasm-pkg-client/src/decoded_component.rs
@@ -222,7 +222,7 @@ fn extract_resolve_and_world_id(
         DecodedWasm::WitPackage(resolve, pkg) => {
             let world_id = resolve
                 .select_world(&[*pkg], None)
-                .map_err(Error::InvalidPackage)?;
+                .map_err(|e| Error::InvalidPackage(e.context("we currently do not support multiple wolds, etc...")))?;
             Ok((resolve, world_id))
         }
     }
--- a/crates/wasm-pkg-client/src/decoded_component.rs
+++ b/crates/wasm-pkg-client/src/decoded_component.rs
@@ -212,10 +212,8 @@ fn extract_package_version(decoded: &DecodedWasm) -> Result<(PackageRef, Version
 /// for semver checking.
 ///
 /// Semver checks compare two versions of the same component, and a
-/// component is defined by exactly one world; thus a `WitPackage` must
-/// define exactly one world.
-///
-/// A multi-world `WitPackage` describes multiple components that must be handled individually.
+/// component is defined by exactly one world. Therefore it is considered
+/// an error if a `WitPackage` defines more than one world.
 fn extract_resolve_and_world_id(
     decoded: &DecodedWasm,
 ) -> Result<(&wit_parser::Resolve, wit_parser::WorldId), Error> {
--- a/crates/wasm-pkg-client/src/decoded_component.rs
+++ b/crates/wasm-pkg-client/src/decoded_component.rs
@@ -208,10 +208,14 @@ fn extract_package_version(decoded: &DecodedWasm) -> Result<(PackageRef, Version
     Ok((package, version))
 }
 
-/// Borrow the inner `wit_parser::Resolve` and resolve a concrete `WorldId`.
-/// For a decoded component the world is fixed; for a WIT package we ask
-/// `Resolve::select_world` to pick one — deferred until needed so a
-/// multi-world WIT package can publish its first version unambiguously.
+/// Borrow the inner `wit_parser::Resolve` and resolve a concrete `WorldId`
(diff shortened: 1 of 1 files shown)


crates/wasm-pkg-client/src/decoded_component.rs
```diff
@@ -208,10 +208,12 @@ fn extract_package_version(decoded: &DecodedWasm) -> Result<(PackageRef, Version
...
-/// Borrow the inner `wit_parser::Resolve` and resolve a concrete `WorldId`.
-/// For a decoded component the world is fixed; for a WIT package we ask
-/// `Resolve::select_world` to pick one — deferred until needed so a
-/// multi-world WIT package can publish its first version unambiguously.
+/// Borrow the inner `wit_parser::Resolve` and resolve a concrete `WorldId`
+/// for semver checking.
+///
+/// Semver checks compare two versions of the same component, and a
+/// component is defined by exactly one world. Therefore it is considered
+/// an error if a `WitPackage` defines more than one world.
```
mkatychev, line 216: Are we saying that a multi world package is unsupported or unexpected/undesired?

ryan-surname-p: I'd say it's more of a limitation of the current implementation because I was working under the incorrect assumption that a published artifact would map 1:1 to a component. 

This is likely to be the common case but it's not strictly true and it will require some additional handling on the client to fix this. 

Namely that we need to decode the WitPackage and extract all the components defined within. 

While I'm quite certain there is a 1:1 mapping of Component to world I'm not sure if there is a 1:1 mapping of world to component. 

Do you happen to know?

Given that you are preparing a release it seemed prudent not to try expand this functionality today but to instead document the existing behavior.

ryan-surname-p: I'm probably getting a bit too hung up on the wording here but "can be thought of" causes some doubt.  

https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md?plain=1#L159C42-L160C57

ryan-surname-p: https://github.com/bytecodealliance/wasm-pkg-tools/issues/230

mkatychev: > While I'm quite certain there is a 1:1 mapping of Component to world I'm not sure if there is a 1:1 mapping of world to component.

You're usually able to map a component to a known world or one that a component creates, you can see this by calling `--importize` in wasm-tools:

```sh-session
$ wasm-tools component wit my_component_0.1.0.wasm --importize
package root:component;

world root-importized {
  ...
```

crates/wasm-pkg-client/src/decoded_component.rs
mkatychev: for now can we add an anyhow context saying multi-world extraction is currently unsupported?

```suggestion
                .map_err(|e| Error::InvalidPackage(e.context("we currently do not support multiple wolds, etc...")))?;
```

crates/wasm-pkg-client/src/decoded_component.rs
mkatychev: I think for now it's safe to go with the assumption that a component will map to a well-known world or a `root:component` it creates for itself.
pinged @salmans to config.

hiSandog:
Would it be useful for the error text to distinguish a package with zero worlds from one with multiple worlds? Both violate the single-world requirement, but the likely corrective action differs: define a world versus select or split the intended component contract. Including the observed world count, and possibly names when there are only a few, could make this clarification more actionable.

crates/wasm-pkg-client/src/decoded_component.rs
salmans: Every component targets a single world (this is the top level world you get when you do `wasm-tools component wit`). I think it's a safe assumption that there's always a single (default) world for `select_world` to pick from, if the source is always a decoded WIT from a compiled component. And components can target the same world (nominally identified by a package, name, and version number). But also notice that in CM, type checking is structural, so `root:component` is just a placeholder that structurally matches the world that the component targets.

ryan-surname-p: After struggling with it a bit more I opened https://github.com/bytecodealliance/wasm-pkg-tools/pull/232 to remove this restriction entirely. 

Thanks to you both I have a much better understanding of how WIT packages and the realization of the types defined within come together. 

Since #232 has been merged into main I'll go ahead and close this PR.

PR #231, opened 2026-07-10
feat(wkg,core,client): intial wkg workspace approach

mkatychev:
PR adds initial workspace support for `wkg fetch` and `wkg publish`

`wkg publish`:
* added `--workspace` flag to select packages for publishing from the `wkg.toml` manifest (if `workspace` table is present)
* added `--skip-dupes`: to allow idempotent publishing, `wkg publish` will skip publishing a version that is already available upstream instead of producing an error.

`wkg fetch`:
* supports multi package lockfiles by unioning the dependency trees of all included packages
* supports workspace fetching

Introduced `WorkspaceConfig` to manage workspace configurations in manifests:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/06160614ccd59b899af74a9daff246282351b857/crates/wasm-pkg-core/src/manifest/workspace.rs#L11-L20

`WorkspaceRootConfig` added as a virtual-manifest stand-in:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/06160614ccd59b899af74a9daff246282351b857/crates/wasm-pkg-core/src/manifest/workspace.rs#L48-L53

### DIFF
--- a/crates/wasm-pkg-core/src/manifest.rs
+++ b/crates/wasm-pkg-core/src/manifest.rs
@@ -123,7 +123,7 @@ impl Manifest {
         let Some(manifest_file) = find_root_manifest_for_wd(cwd) else {
             return Ok(None);
         };
-        let manifest_dir = manifest_file.parent().unwrap();
+        let manifest_dir = manifest_file.parent().context("unexpectedly missing directory containing manifest")?;
         let manifest = Self::load_from_path(&manifest_file).await?;
 
         if let Some(root) = manifest.root() {
--- a/crates/wasm-pkg-core/src/manifest.rs
+++ b/crates/wasm-pkg-core/src/manifest.rs
@@ -10,14 +10,28 @@ use semver::VersionReq;
 use serde::{Deserialize, Serialize};
 use tokio::io::AsyncWriteExt;
 
+mod paths;
+pub mod workspace;
+
+use workspace::*;
+
+use crate::manifest::paths::{find_root_iter, find_root_manifest_for_wd};
+
 /// The default name of the manifest file.
 pub const MANIFEST_FILE_NAME: &str = "wkg.toml";
+/// Directory next to the root [`MANIFEST_FILE_NAME`] that holds multi-package `deps` and `config.toml`.
+pub const WORKSPACE_OUT_DIR: &str = "wkg";
 
 /// The structure for a wkg.toml manifest file. This file is entirely optional and is used for
 /// overriding and annotating wasm packages.
+/// `workspace` is mutually exclusive with `overrides` and top-level `metadata`
 #[derive(Debug, Clone, Serialize, Deserialize, Default, PartialEq, Eq)]
 #[serde(deny_unknown_fields)]
 pub struct Manifest {
+    /// Workspace declaration.
+    // TODO: this should be a `TomlWorkspace` so that serialization is not coupled to the config structure
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub workspace: Option<WorkspaceConfig>,
     /// Overrides for various packages
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub overrides: Option<HashMap<String, Override>>,
@@ -28,15 +42,68 @@ pub struct Manifest {
 }
 
 impl Manifest {
+    fn from_toml(contents: &str) -> Result<Manifest> {
+        let manifest: Manifest =
+            toml::from_str(contents).context("unable to parse manifest file")?;
+        manifest.validate()?;
+        Ok(manifest)
+    }
+
     /// Loads a manifest file from the given path.
     pub async fn load_from_path(path: impl AsRef<Path>) -> Result<Manifest> {
-        tracing::info!(path = %path.as_ref().display(), "loading wkg manifest file");
-        let contents = tokio::fs::read_to_string(path)
-            .await
-            .context("unable to load manifest from file")?;
-        let manifest: Manifest =
-            toml::from_str(&contents).context("unable to parse manifest file")?;
+        let path = path.as_ref();
+        tracing::info!(path = %path.display(), "loading wkg manifest file");
+        let contents = std::fs::read_to_string(path)
+            .with_context(|| format!("unable to load manifest from {}", path.display()))?;
+        let mut manifest = Self::from_toml(&contents)
+            .with_context(|| format!("invalid manifest at {}", path.display()))?;
+        if let Some(WorkspaceConfig::Root(root)) = &mut manifest.workspace {
+            root.root_dir = path
+                .parent()
+                .with_context(|| {
+                    format!("manifest path has no parent directory: {}", path.display())
+                })?
+                .to_path_buf();
+            // Resolve globs and relative paths eagerly
+            root.members = WorkspaceRootConfig::resolve_members(&root.members, &root.root_dir);
+        }
         Ok(manifest)
+    }
+
+    fn root(&self) -> Option<&WorkspaceRootConfig> {
+        if let Some(WorkspaceConfig::Root(root)) = &self.workspace {
+            return Some(&root);
+        }
+        None
+    }
+
+    // `Manifest` validations, mirrors cargo's `Workspace::validate`
+    fn validate(&self) -> Result<()> {
+        self.validate_workspace_exclusivity()?;
+        // Add new validation rules with `self.validate_*()?;`
+        Ok(())
+    }
+
+    // no overrides or top-level metadata when workspace is present
+    fn validate_workspace_exclusivity(&self) -> Result<()> {
+        if self.workspace.is_none() {
+            return Ok(());
+        }
+        let mut conflicts = Vec::new();
+        if self.overrides.is_some() {
+            conflicts.push("overrides");
+        }
+        if self.metadata.is_some() {
+            conflicts.push("metadata");
+        }
+        if conflicts.is_empty() {
+            return Ok(());
+        }
+        anyhow::bail!(
+            "`[workspace]` cannot coexist with: `[{}]` - \
+             use `[workspace.metadata]` for workspace level values",
+            conflicts.join("]`, `[")
+        );
     }
 
     /// Attempts to load the manifest from the current directory. Most of the time, users of this
@@ -49,6 +116,32 @@ impl Manifest {
             return Ok(Manifest::default());
         }
         Self::load_from_path(manifest_path).await
+    }
+
+    /// Tries to find the root workspace config
+    /// Returns `Ok(None)` when there is no `wkg.toml` ancestor that can be [`WorkspaceRootConfig`]
+    pub async fn load_root_workspace(cwd: &Path) -> Result<Option<WorkspaceRootConfig>> {
+        let Some(manifest_file) = find_root_manifest_for_wd(cwd) else {
+            return Ok(None);
+        };
+        let manifest_dir = manifest_file.parent().unwrap();
+        let manifest = Self::load_from_path(&manifest_file).await?;
+
+        if let Some(root) = manifest.root() {
+            return Ok(Some(root.clone()));
+        }
+
+        // keep walking up if we have not found root
+        for file in find_root_iter(&manifest_file) {
+            let manifest = Self::load_from_path(&file).await?;
+            if let Some(WorkspaceConfig::Root(root)) = manifest.workspace {
+                if root.is_explicitly_listed_member(&manifest_dir) {
+                    return Ok(Some(root));
+                }
+            }
+        }
+
+        Ok(None)
     }
 
     /// Serializes and writes the manifest to the given path.
@@ -117,6 +210,7 @@ mod tests {
         let tempdir = tempfile::tempdir().unwrap();
         let manifest_path = tempdir.path().join(MANIFEST_FILE_NAME);
         let manifest = Manifest {
+            workspace: None,
             overrides: Some(HashMap::from([(
                 "foo:bar".to_string(),
                 Override {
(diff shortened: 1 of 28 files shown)


crates/wasm-pkg-core/src/manifest/workspace.rs
```diff
@@ -0,0 +1,311 @@
...
+    ///
+    /// Returns `true` ONLY if:
+    /// - The path is the workspace root manifest itself, or
+    /// - The path matches one of the explicit `members` patterns
+    ///
+    /// NOTE: This does NOT check for implicit path dependency membership.
+    /// A `false` return does NOT mean the package is definitely not a member -
+    /// it could still be a member via path dependencies. Callers should fallback
+    /// to full workspace loading when this returns `false`.
+    // FIXME: implement cargo WorkspaceRootConfig::{member_paths, expand_member_path}
```
mkatychev, line 68: Current workspace members are eagerly evaluated to this functions more like `is_member` for the moment, I can implement the `member_paths` and `expand_member_path` functionality here or in a follow-up. The distinction should not matter until `WorkspaceConfig::Member` is implemented.

crates/wasm-pkg-core/src/manifest.rs
```diff
@@ -51,6 +118,33 @@ impl Manifest {
         Self::load_from_path(manifest_path).await
     }
 
+    /// Tries to find the root workspace config
+    /// Returns `Ok(None)` when there is no `wkg.toml` ancestor that can be [`WorkspaceRootConfig`]
+    // TODO(maktychev): reconcile load_from_path and load_root_workspace
+    pub async fn load_root_workspace(cwd: &Path) -> Result<Option<WorkspaceRootConfig>> {
```
mkatychev, line 121: ideally this should use `&mut self` [similar to cargo](https://github.com/rust-lang/cargo/blob/a595d0da21f228b7fdae64d3d5c0e527ea66bb59/src/cargo/core/workspace.rs#L803-L821) but for now we don't have a `Workspace` struct

crates/wasm-pkg-core/src/manifest/paths.rs
```diff
@@ -0,0 +1,102 @@
+use anyhow::Result;
+use std::path::{Component, Path, PathBuf};
+
+use crate::manifest::MANIFEST_FILE_NAME;
+
+/// Find the first ancestor [`super::Manifest`] path for current working directory.
```
vados-cosmonic, line 6: IMO this function isn't about the working directory -- it really works on any directory, so maybe `find_root_maniest_for_dir` is more appropriate?

crates/wasm-pkg-core/src/manifest/workspace.rs
```diff
@@ -0,0 +1,306 @@
...
+    Root(WorkspaceRootConfig),
+    // - `workspace = "path/to/workspace/root"`
+    // - TODO(mkatychev): implement workspace package level field
+    // - https://doc.rust-lang.org/cargo/reference/manifest.html#the-workspace-field
+    // Member {
+    //     root: Option<String>,
+    // },
+}
+
+impl Default for WorkspaceConfig {
```
vados-cosmonic, line 18: NIT: We can `#[derive(Default)]` and `#[default]` for this IMO

crates/wasm-pkg-core/src/manifest/workspace.rs
```diff
@@ -0,0 +1,306 @@
...
+
+use super::*;
+use std::path::{Path, PathBuf};
+
+use serde::{Deserialize, Serialize};
+
+/// `[workspace]` table inside a `wkg.toml` manifest.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum WorkspaceConfig {
```
vados-cosmonic, line 11: What do you think about this being `non_exhaustive`?, it's almost guaranteed to change in the future (at the very least we have actual plans to change this)

I'm of two minds because we're 0.x and we can just release a breaking change when we need to.

crates/wasm-pkg-core/src/manifest/workspace.rs
vados-cosmonic: IMO let's do it in a follow up

crates/wasm-pkg-core/src/manifest/workspace.rs
```diff
@@ -0,0 +1,306 @@
...
+        assert!(find_root_manifest_for_wd(tempdir.path()).is_none());
+    }
+
+    #[test]
+    fn find_root_iter_ancestors() {
+        // dir
+        // ├── wkg.toml
+        // └── pkg
+        //     └── wkg.toml
+        let tempdir = tempfile::tempdir().unwrap();
```
vados-cosmonic, line 131: NIT: I really wonder if these should be fixtures. Feels wasteful to add like 10 new folders, but also feels unfortunate to write temp files in the unit tests.

crates/wasm-pkg-core/src/manifest.rs
```diff
@@ -28,17 +40,70 @@ pub struct Manifest {
 }
 
 impl Manifest {
+    fn from_toml(contents: &str) -> Result<Manifest> {
+        let manifest: Manifest =
+            toml::from_str(contents).context("unable to parse manifest file")?;
```
vados-cosmonic: I think you might not need this, because this context is better served from where you know the file that failed to parse, and you do that below

crates/wasm-pkg-core/src/manifest.rs
```diff
@@ -51,11 +116,36 @@ impl Manifest {
...
     }
 
+    /// Tries to find the root workspace config
+    /// Returns `Ok(None)` when there is no `wkg.toml` ancestor that can be [`WorkspaceRootConfig`]
+    // TODO(maktychev): reconcile load_from_path and load_root_workspace
+    pub async fn load_root_workspace(cwd: &Path) -> Result<Option<WorkspaceRootConfig>> {
+        let Some(manifest_file) = find_root_manifest_for_wd(cwd) else {
+            return Ok(None);
+        };
+        let manifest_dir = manifest_file.parent().unwrap();
```
vados-cosmonic:
```suggestion
        let manifest_dir = manifest_file.parent().context("unexpectedly missing directory containing manifest")?;
```

crates/wasm-pkg-core/src/manifest.rs
```diff
@@ -51,11 +116,36 @@ impl Manifest {
...
+        let manifest = Self::load_from_path(&manifest_file).await?;
+
+        if let Some(root) = manifest.root() {
+            return Ok(Some(root.clone()));
+        }
+
+        // keep walking up if we have not found root
+        for file in find_root_iter(&manifest_file) {
```
vados-cosmonic, line 135: wonder if this should just be rolled into the `find_root` fn? `find_root_from_manifest`? It seems like this is never called by itself

crates/wkg/src/wit.rs
```diff
@@ -150,24 +154,165 @@ pub async fn temp_wit_file(package: &PackageRef, bytes: &[u8]) -> anyhow::Result
...
 impl FetchArgs {
     pub async fn run(self) -> anyhow::Result<()> {
-        check_dir(&self.dir).await?;
-        let client = self.common.get_client().await?;
-        let manifest = Manifest::load().await?;
-        let mut lock_file = LockFile::load(false).await?;
-        wit::fetch_dependencies(
-            &manifest,
-            self.dir,
+        let cwd = std::env::current_dir()?;
```
vados-cosmonic, line 157: What do you think abou workspace root dir as a property w/ a default to current dir when it's missing?

vados-cosmonic (approved):
LGTM 🚀 

Mostly nits, and feel free to push any you want to followup(s)!

crates/wasm-pkg-core/src/manifest/paths.rs
mkatychev: I copied it over from https://github.com/rust-lang/cargo/blob/24e0f7dc0d5adeffb21929740c2b43664230e185/src/cargo/util/important_paths.rs#L6 😅 
I think for now, given how much forgery I am doing with cargo structs/methods - I vote keeping it the same as a matter of reference until the workspace is fleshed out.

crates/wasm-pkg-core/src/manifest/workspace.rs
mkatychev: Agreed, should I uncomment the member variant then?

mkatychev: At the very least with `Self::Member` I would consider it exhaustive, what about adding an `expect(dead_code)` above `Self::Member`?

crates/wasm-pkg-core/src/manifest/workspace.rs
mkatychev: These are all temp folders so they are cleaned up on `tempfile::Tempdir` drop and never committed.

crates/wkg/src/wit.rs
mkatychev: a property for `FetchArgs`? Makes sense.

crates/wasm-pkg-core/src/manifest/workspace.rs
mkatychev: Oh yeah, that's not possible yet:
```
rustc: the `#[default]` attribute may only be used on unit enum variants
consider a manual implementation of `Default`
```

crates/wasm-pkg-core/src/manifest.rs
mkatychev: Are you suggesting inlining into `load_from_path`? It's called mostly in tests and here:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/e3d72524122ec47517e8df3874f15eafffe3668a/crates/wasm-pkg-core/src/manifest.rs#L56-L57

Given that the function is private, should we add a `cfg(test)` above it and inline the call inside `load_from_path`?

crates/wasm-pkg-core/src/manifest.rs
mkatychev: This is another one of the cargo holdovers:

https://github.com/rust-lang/cargo/blob/a595d0da21f228b7fdae64d3d5c0e527ea66bb59/src/cargo/core/workspace.rs#L2286-L2306

https://github.com/rust-lang/cargo/blob/a595d0da21f228b7fdae64d3d5c0e527ea66bb59/src/cargo/core/workspace.rs#L2255-L2272

Once we have a `GlobalContext`, it might make more sense to keep it as-is, thoughts?

crates/wkg/src/wit.rs
mkatychev: `cargo fetch` only has a manifest path arg:
```
--manifest-path <PATH>  Path to Cargo.toml
```

And `cargo` proper has a nightly-only chdir arg:
```
  -C <DIRECTORY>                 Change to DIRECTORY before doing anything (nightly-only)
```

crates/wasm-pkg-core/src/manifest/workspace.rs
vados-cosmonic: So I meant to actually avoid doing the read and write *during* the test

crates/wasm-pkg-core/src/manifest.rs
vados-cosmonic: Oh I meant that  literally the callls to `.context` isnide `toml::from_str` were probably enough

crates/wasm-pkg-core/src/manifest.rs
vados-cosmonic: Ah OK, if you have this in preparation for that, sure!

crates/wkg/src/wit.rs
vados-cosmonic: Hmnn yeah, if the answer is that the user should always run from the right folder then that's absolutely fine too! Just want to make sure this is an explicit decision. 

Dunno if I like the `chdir` because that's not quite what we're doing here BUT if you feel like that's a better mapping than allowing the option,, sure!

crates/wasm-pkg-core/src/manifest/workspace.rs
vados-cosmonic: Clearly we have lots of different cases here, but none of them actually change (and deserve new files being written), which is why I was thinking they could be fixtures

crates/wkg/src/wit.rs
mkatychev: So what was your original suggestion? `--manifest-path` makes the most sense (if we use the directory relative to the manifest provided), I feel like a dir argument can be solved with a `(cd /my_dir/ && wkg fetch)`

crates/wasm-pkg-core/src/manifest/workspace.rs
mkatychev: I see, so we do the dir/file creation during the test but as fixture function similar to the changes I did in e2e for workspace fixtures:
https://github.com/bytecodealliance/wasm-pkg-tools/pull/231/files#diff-ab2d8c35c52a8931fc6d105f132b9fc37d09a88b215b94883197747d9ebda1cf

I feel that the file tree comments are still nice, it was the easiest way for me to visualize the test case in my mind.

mkatychev: I agree with you on this, I would like to defer this to mirror cargo's `#[cargo_test]` approach here:
https://github.com/rust-lang/cargo/blob/a595d0da21f228b7fdae64d3d5c0e527ea66bb59/tests/testsuite/workspaces.rs#L2867-L2932

crates/wasm-pkg-core/src/manifest/workspace.rs
vados-cosmonic: Ah I don't think you have to add the member variant yet! I'm just saying that we can leave that off to the future if we mark non_exhaustive now!

mkatychev: It was easier just to add it in, everything compiled 🤷

PR #232, opened 2026-07-10
semver checking for multi-world wit packages

### LINKED ISSUE #230: semver check should support multi-world wit packages
There is currently no mechanism in place to break a multi-world WitPackage into the multiple Components they represent. 

As a consequence we are not able to perform semver verification against multi-world WithPackages.

ryan-surname-p:
closes #230 

Extracts each world id and its world name from `WitPackage` then compares each as it would a Component. 

Assumes adding a world to a package is a compatible change while removing one is a breaking change.

### DIFF
--- a/crates/wasm-pkg-client/src/decoded_component.rs
+++ b/crates/wasm-pkg-client/src/decoded_component.rs
@@ -88,34 +88,47 @@ impl DecodedComponent {
             (other, self)
         };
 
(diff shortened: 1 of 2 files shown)


mkatychev (commented):
This looks great, thanks for getting fix out so fast, I will take a better look once back.

mkatychev (approved):
Actually, this is a solid fix for the gap, LGTM

PR #237, opened 2026-07-14
feat(config,client): provide a per-registry envvar scheme to set autentication

mkatychev:
This fills a gap where the only current ways of overriding configs through envvars are only through the `wkg oci` subcommands with `WKG_OCI_USERNAME` and `WKG_OCI_PASSWORD`:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/6ca71800a5f7733c2276d16ce7ba679769468f9f/crates/wkg/src/oci.rs#L14-L32 

Many OCI registries do not expect a username/password scheme and instead want to be given a bearer token.

This attempts to fill the gap by exposing `WKG_REGISTRY_<REGISTRY>_AUTH_<AUTH_SCHEME>` as a means of providing credentials for a given registry.

This attempts to account for scenarios where multiple registries are authenticated with at once.

### DIFF
--- a/crates/wasm-pkg-client/src/oci/mod.rs
+++ b/crates/wasm-pkg-client/src/oci/mod.rs
@@ -114,6 +114,13 @@ impl OciBackend {
     }
 
     pub(crate) fn get_credentials(&self) -> Result<RegistryAuth, Error> {
+        // Prefer a bearer token from envvar if it exists
+        if let Ok(token) = std::env::var(registry_token_env_var(&self.oci_registry))
+            && !token.is_empty()
+        {
+            return Ok(RegistryAuth::Bearer(token));
+        }
+
         if let Some(BasicCredentials { username, password }) = &self.credentials {
             return Ok(RegistryAuth::Basic(
                 username.clone(),
@@ -176,6 +183,22 @@ pub(crate) fn oci_registry_error(err: OciDistributionError) -> Error {
     }
 }
 
+/// Returns the envvar used to provide a bearer token for a given registry:
+/// `WKG_REGISTRIES_<name>_TOKEN`, where registry `<name>` is defined by
+/// the resolution of [`wasm_pkg_common::config::Config`].
+///
+/// Any non-ASCII-alphanumeric characters (., -, :, /, ...) are replaced by `_` and
+/// the whole string is upper-cased.
+// TODO(mkatychev): move this into `wasm_pkg_common::config::Config` once overlays are implemented,
+// this should be a generic-to-backend way of overriding configs.
+fn registry_token_env_var(oci_registry: &str) -> String {
+    let sanitized: String = oci_registry
+        .chars()
+        .map(|c| if c.is_ascii_alphanumeric() { c } else { '_' })
+        .collect();
+    format!("WKG_REGISTRIES_{}_TOKEN", sanitized.to_ascii_uppercase())
+}
+
 fn get_docker_credential(registry: &str) -> Result<Option<RegistryAuth>, Error> {
     match docker_credential::get_credential(registry) {
         Ok(DockerCredential::UsernamePassword(username, password)) => {
@@ -202,4 +225,25 @@ fn get_docker_credential(registry: &str) -> Result<Option<RegistryAuth>, Error> 
     }
 
     Ok(None)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn token_env_var_sanitizes_registry_name() {
+        assert_eq!(
+            registry_token_env_var("example.com"),
+            "WKG_REGISTRIES_EXAMPLE_COM_TOKEN"
+        );
+        assert_eq!(
+            registry_token_env_var("ghcr.io"),
+            "WKG_REGISTRIES_GHCR_IO_TOKEN"
+        );
+        assert_eq!(
+            registry_token_env_var("localhost:1234"),
+            "WKG_REGISTRIES_LOCALHOST_1234_TOKEN",
+        );
+    }
 }
--- a/crates/wasm-pkg-client/src/oci/mod.rs
+++ b/crates/wasm-pkg-client/src/oci/mod.rs
@@ -114,10 +114,13 @@ impl OciBackend {
     }
 
     pub(crate) fn get_credentials(&self) -> Result<RegistryAuth, Error> {
-        // Prefer a bearer token from envvar if it exists
-        if let Ok(token) = std::env::var(registry_token_env_var(&self.oci_registry))
+        // Detect `WKG_REGISTRY_<REGISTRY>_AUTH_<AUTH_SCHEME>` if present.
+        let auth_var_key = &registry_auth_env_var(&self.oci_registry, "BEARER");
+        if let Ok(token) = std::env::var(auth_var_key)
             && !token.is_empty()
         {
+            tracing::debug!(registry = %self.oci_registry, %auth_var_key, "Using detected authentication envvar key");
+            // Only `BEARER` AUTH_SCHEME for now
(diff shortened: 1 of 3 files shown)


mkatychev:
@vados-cosmonic @ryan-surname-p I know this is a bit "hacky" so I've included this PR as a draft/proposal functionality.

Ideally I'd like to see this land in 0.16 (with further tests & discussion of course) because there is a real gap in configuration currently where CI has to explicitly override config files with bearer tokens because there is no means for doing so from the environment

I usually end up having to overwrite/create files and merge them with existing configs that have password as a placeholder:
```nushell
# wkg-config set-password --registry my-registry --password $env.OCI_BEARER_TOKEN config.toml
export def set-password [--registry: string --password: string path: path] {
    # override the config.toml path
    open $path 
    | merge deep -s overwrite {registry: {$registry: {oci: { auth: { password: $"($password)"}}}}}
    | save --force $path
}

def main [] { }
```


This change should not have regressions as we do not have an existing flow for bearer tokens

ryan-surname-p:
> @vados-cosmonic @ryan-surname-p I know this is a bit "hacky" so I've included this PR as a draft/proposal functionality.
> 
> Ideally I'd like to see this land in 0.16 (with further tests & discussion of course) because there is a real gap in configuration currently where CI has to explicitly override config files with bearer tokens because there is no means for doing so from the environment
> 
> I usually end up having to overwrite/create files and merge them with existing configs that have password as a placeholder:
> 
> ```nushell
> # wkg-config set-password --registry my-registry --password $env.OCI_BEARER_TOKEN config.toml
> export def set-password [--registry: string --password: string path: path] {
>     # override the config.toml path
>     open $path 
>     | merge deep -s overwrite {registry: {$registry: {oci: { auth: { password: $"($password)"}}}}}
>     | save --force $path
> }
> 
> def main [] { }
> ```
> 
> This change should not have regressions as we do not have an existing flow for bearer tokens

---

This seems reasonable for mitigating the issue in the near term.

I'm partial to the overlay approach which it looks like you have in the plans already anyway baed on the todo.

crates/wasm-pkg-client/src/oci/mod.rs
```diff
@@ -176,6 +183,22 @@ pub(crate) fn oci_registry_error(err: OciDistributionError) -> Error {
...
+/// Any non-ASCII-alphanumeric characters (., -, :, /, ...) are replaced by `_` and
+/// the whole string is upper-cased.
+// TODO(mkatychev): move this into `wasm_pkg_common::config::Config` once overlays are implemented,
+// this should be a generic-to-backend way of overriding configs.
+fn registry_token_env_var(oci_registry: &str) -> String {
+    let sanitized: String = oci_registry
+        .chars()
+        .map(|c| if c.is_ascii_alphanumeric() { c } else { '_' })
+        .collect();
+    format!("WKG_REGISTRIES_{}_TOKEN", sanitized.to_ascii_uppercase())
```
vados-cosmonic:
```suggestion
    format!("WKG_REGISTRY_{}_AUTH_BEARER_TOKEN", sanitized.to_ascii_uppercase())
```

crates/wasm-pkg-client/src/oci/mod.rs
```diff
@@ -176,6 +183,22 @@ pub(crate) fn oci_registry_error(err: OciDistributionError) -> Error {
...
+/// Returns the envvar used to provide a bearer token for a given registry:
+/// `WKG_REGISTRIES_<name>_TOKEN`, where registry `<name>` is defined by
+/// the resolution of [`wasm_pkg_common::config::Config`].
+///
+/// Any non-ASCII-alphanumeric characters (., -, :, /, ...) are replaced by `_` and
+/// the whole string is upper-cased.
+// TODO(mkatychev): move this into `wasm_pkg_common::config::Config` once overlays are implemented,
+// this should be a generic-to-backend way of overriding configs.
+fn registry_token_env_var(oci_registry: &str) -> String {
+    let sanitized: String = oci_registry
```
vados-cosmonic, line 199: What do you think about actually *not* modifying the registry URL at all? What are we avoiding here?

I don't think there's much wrong with `WKG_REGISTRY_example.com_AUTH_BEARER_TOKEN`...

crates/wasm-pkg-client/src/oci/mod.rs
```diff
@@ -114,6 +114,13 @@ impl OciBackend {
     }
 
     pub(crate) fn get_credentials(&self) -> Result<RegistryAuth, Error> {
+        // Prefer a bearer token from envvar if it exists
+        if let Ok(token) = std::env::var(registry_token_env_var(&self.oci_registry))
+            && !token.is_empty()
+        {
+            return Ok(RegistryAuth::Bearer(token));
```
vados-cosmonic, line 124: What do you think about leaving a log message (`tracing::debug`?) here to make it clear we took this shortcut/the ENV was honored?

vados-cosmonic (approved):
Looks good to me -- happy to have this to make things slightly easier for actual use.

crates/wasm-pkg-client/src/oci/mod.rs
mkatychev: Most shells don't support it 🤷:
```sh-session
sh-5.3$ WKG_REGISTRY_example.com_AUTH_BEARER_TOKEN="foo"
sh: WKG_REGISTRY_example.com_AUTH_BEARER_TOKEN=foo: command not found
```

vados-cosmonic: Ah OK, yep that's a problem -- thanks for pointing that out. I guess there's not much choice here. The idea of `:` and `-` both going to `_` is a bit weird to me.

Alternatively something like 
```
WKG_REGISTRIES_AUTH_BEARER_TOKENS=example.com=<....>,somewhere-else.io=<...>
```
Seems a bit lengthy. 

Yeah I'm fine with the replacements as is then.

mkatychev: yeah, I'd rather defer mangling until we actually start dealing with collisions 😅. Not that it won't happen... but just that I do not want to deal with it at the moment.

crates/wasm-pkg-client/src/oci/mod.rs
mkatychev: Do you feel that's too wordy? `WKG_REGISTRY_GHCR_IO_AUTH_BEARER_TOKEN="ey......."`?

vados-cosmonic: Yeah... maybe? I think the two things I was thinking were:
* this applies to one registry (not multiple)
* "token" doesn't really say much, at the very least it should be immediately obvious that it's an auth token... But then again PATs are token

vados-cosmonic: I don't feel that strongly about this on second thought -- especially if it's a sort of temporary fix... Then again these kind of temporary fixes are the kind that stay around a long time and get depended on... I'm actually not sure I'd want to take it away

ryan-surname-p: If we drop "TOKEN" from the template we have `WKG_REGISTRY_{REGISTRY}_AUTH_{AUTH_SCHEME}` which can initially just accept "BEARER" but can evolve later without too much fuss.

mkatychev: I like that

crates/wasm-pkg-client/src/oci/mod.rs
ryan-surname-p: Without too much effort we could probably do a quick registry query and raise an error if we are unable to disambiguate the result.

ryan-surname-p (approved):

PR #239, opened 2026-07-27
fix(oci): handle title metadata

### LINKED ISSUE #238: wkg packages publishes/pushed to OCI missing org.opencontainers.image.title for layers
### Description

Artifacts pushed by `wkg publish` and `wkg oci push` don't include the `org.opencontainers.image.title` annotation on their application/wasm layers. OCI tools like `oras` will skip over pulling the layers missing this annotation that they use for the filename on disk.

### Repro

```
wkg oci push registry.example.com/myns/mypkg:1.0.0 mypkg.wasm
oras pull registry.example.com/myns/mypkg:1.0.0 --output ./out
# exits 0, ./out is empty
```

### Fix

wkg includes the `org.opencontainers.image.title` the layers defined in the package manifest, e.g.:

```json
{
  "annotations": {
      "org.opencontainers.image.version": "1.2.3"
  },
  "config": {
      "digest": "sha256:...",
      "mediaType": "application/vnd.wasm.config.v0+json",
      "size": 123
  },
  "layers": [
      {
          "digest": "sha256:...",
          "mediaType": "application/wasm",
          "size": 12345,
          "annotations": {
              "org.opencontainers.image.title": "mypkg.wasm"
          }
      }
  ],
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "schemaVersion": 2
}
```

mkatychev:
Resolves https://github.com/bytecodealliance/wasm-pkg-tools/issues/238

[`org.opencontainers.image.title`](https://specs.opencontainers.org/image-spec/annotations/#pre-defined-annotation-keys) was previously not being set for `wkg publish` or `wkg oci push` causing certain OCI tools to be unable to pull the components.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1864,9 +1864,9 @@ dependencies = [
 
 [[package]]
 name = "oci-wasm"
-version = "0.5.0"
+version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "efc8835b4a949024f418e8a7f25db76038da0d3d670a1507a6e346dd06ae7ccb"
+checksum = "87689298bd74f0f2675fcac99956a34c31098ca3bdced3d7635e71dc03c5ca21"
 dependencies = [
  "anyhow",
  "chrono",
@@ -1875,8 +1875,8 @@ dependencies = [
  "serde_json",
  "sha2 0.11.0",
  "tokio",
- "wit-component 0.252.0",
- "wit-parser 0.252.0",
+ "wit-component",
+ "wit-parser",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


vados-cosmonic (approved):
LGTM 🚀

vados-cosmonic:
Looks like the CI test is a flake -- probably should pass if you re-run? I guess there's a small chance that docker does not support this annotation, in which case we might have to make this an option

mkatychev:
> Looks like the CI test is a flake -- probably should pass if you re-run? I guess there's a small chance that docker does not support this annotation, in which case we might have to make this an option

good point, we'll have to see, this has definitely caused issues with CLI tools like `oras` (have confirmed it).

PR #241, opened 2026-07-27
doc: update with workspace and other 0.16 changes

mkatychev:
This PR adds changes for the root cert override and workspace additions to the documentation.


mkatychev:
Merging now for 0.16.0 release but will address feedback

PR #244, opened 2026-07-30
fix(wkg): respect member manifests during fetch

### LINKED ISSUE #243: bug: `wkg fetch` will attempt workspace fetching even if a member manifest is present
https://github.com/bytecodealliance/wasm-pkg-tools/blob/dfa2e17128d1b938019d86711f16fb033fddb4f4/crates/wkg/src/wit.rs#L180-L186

This will always call `run_workspace_fetch` if one can path to an ancestor manifest that is a workspace.

subotac:
Fixes #243.

  When `wkg fetch` ran below a workspace root, it always selected workspace mode even when the current directory had its own `wkg.toml`. Use the member manifest as the fetch context in that case while preserving
  workspace behavior when no local manifest exists.

  The regression test verifies that fetching from a member creates the member lock file rather than a workspace lock file.

  ## Testing

  - `cargo test -p wkg --test e2e fetch_from_member_manifest_uses_member_context --no-default-features`
  - `cargo check -p wkg --tests --no-default-features`
  - `cargo test -p wkg --bin wkg --no-default-features`
  - `cargo fmt --all -- --check`


crates/wkg/src/wit.rs
```diff
@@ -155,7 +155,12 @@ pub async fn temp_wit_file(package: &PackageRef, bytes: &[u8]) -> anyhow::Result
 impl FetchArgs {
     pub async fn run(self) -> anyhow::Result<()> {
         let cwd = std::env::current_dir()?;
-        let root = Manifest::load_root_workspace(&cwd).await?;
+        let mut root = Manifest::load_root_workspace(&cwd).await?;
+        if root.as_ref().is_some_and(|root| root.root_dir() != cwd)
+            && tokio::fs::try_exists(cwd.join(MANIFEST_FILE_NAME)).await?
```
mkatychev: Would this approach work if one is in a subdirectory of a member manfiest, for instance, will this work if we are in `subdir`?

```
./root
├── member
│   ├── subdir
│   │   └── file.wit
│   └── wkg.toml
└── wkg.toml
```
`Manifest::load_root_workspace` will keep walking up the path until it finds a manifest that is a workspace root.
It may be better to handle this using `find_root_manifest_for_wd`:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/3d79978ba87c8da69c1dc823645da55e5c41835a/crates/wasm-pkg-core/src/manifest/paths.rs#L6-L7

subotac: Updated this to find the nearest member manifest from nested working directories using `find_root_manifest_for_wd`. The regression now runs from `member/wit/subdir` and verifies that the lock file is created next to the member manifest.

mkatychev (approved):
LGTM

PR #250, opened 2026-08-11
fix(client): apply the package override before deriving identity

salmans:
`DecodedComponent::from_publishing_source_with_package` extracted `(package, version)` from the decoded binary first and applied the override second, so `publish_release_data` failed for any binary without an embedded version before the override was consulted. A WIT package without a version that was published with `PublishOpts { package: Some(..) }` always failed with "version not found", even though that error's help text suggests supplying the identity explicitly.

The change derives the identity from the binary only when no override is supplied. The change left the no-override constructor without callers, so the two constructors are collapsed into one that takes `Option<(PackageRef, Version)>`; the module is crate-private, so there's no public API change.

### DIFF
--- a/crates/wasm-pkg-client/src/decoded_component.rs
+++ b/crates/wasm-pkg-client/src/decoded_component.rs
@@ -16,11 +16,18 @@ pub struct DecodedComponent {
 }
 
 impl DecodedComponent {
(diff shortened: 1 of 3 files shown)


mkatychev (approved):
LGTM, this will have to go in once 0.16.1 is released since it's a breaking change in the API

PR #251, opened 2026-08-15
feat(core,wkg): resolve multiple versions of the same package

Aditya1404Sal:
## Problem

According to the rules governing wit, there's nothing that actively restricts a scenario in which 
a wit world can name more than one version of the same package, say for example when a world exports both

`wasmcloud:messaging/handler@0.2.0` and `@0.3.0` in the same component (from where I ran into this limitation wasmCloud/wasmCloud#5459)

 If there's justification needed as to why this is needed, off the top of my head -- it can help in safe migration
to a new interface rev without a breaking change
or in my use, so that a component can A/B test the new one for regressions
`wit-bindgen` does support it by producing versioned bindings (I think that's what they're called) `messaging0_2_0` and `messaging0_3_0` so I don't see why wkg can't have it in the override resolution

back to the problem : 

Our `wash build` command that builds our components has `WkgFetcher`
which could not resolve the mixed deps and while running the fetch flow, one version was always dropped, and the fetch then failed and cleared the `deps/` sub-dir.

## Change

The resolver is now keyed by `PackageSpec` -- a package plus the version requirement it was
requested under. `[overrides]` keys accept an exact version suffix:

```toml
[overrides]
"my:dep@0.1.0" = { path = "../dep-0.1.0/wit" }
"my:dep@0.2.0" = { path = "../dep-0.2.0/wit" }
```

### existing behaviour

A bare key like below

```toml
[overrides]
"my:dep" = { path = "../dep-0.1.0/wit" }
"my:second_dep" = { path = "../second-dep/wit" }
```

will behave exactly as before.

### DIFF
--- a/crates/wasm-pkg-core/src/manifest.rs
+++ b/crates/wasm-pkg-core/src/manifest.rs
@@ -1,13 +1,14 @@
 //! Type definitions and functions for working with `wkg.toml` files.
 
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     path::{Path, PathBuf},
 };
 
 use anyhow::{Context, Result};
 use semver::VersionReq;
 use serde::{Deserialize, Serialize};
+use wasm_pkg_common::package::{PackageRef, PackageSpec};
 mod paths;
 pub mod workspace;
 
@@ -79,8 +80,54 @@ impl Manifest {
     // `Manifest` validations, mirrors cargo's `Workspace::validate`
     fn validate(&self) -> Result<()> {
         self.validate_workspace_exclusivity()?;
+        self.validate_override_keys()?;
         // Add new validation rules with `self.validate_*()?;`
         Ok(())
+    }
+
+    /// Checks that override keys parse and that no package is covered by both a bare and a
+    /// versioned key.
+    ///
+    /// Runs when a `wkg.toml` is loaded (see [`validate`](Self::validate)), and again when
+    /// resolving, since a `Manifest` built directly in Rust code skips the load step.
+    pub(crate) fn validate_override_keys(&self) -> Result<()> {
+        let Some(overrides) = self.overrides.as_ref() else {
+            return Ok(());
+        };
+        // `overrides` is a map, so walk it in a stable order
+        let mut sorted_keys: Vec<&String> = overrides.keys().collect();
+        sorted_keys.sort_unstable();
+
+        let mut bare: HashSet<PackageRef> = HashSet::new();
+        let mut versioned: HashMap<PackageRef, Vec<&str>> = HashMap::new();
+        for key in sorted_keys {
+            let spec: PackageSpec = key
+                .parse()
+                .with_context(|| format!("invalid override key `{key}`"))?;
+            match spec.version {
+                Some(_) => versioned.entry(spec.package).or_default().push(key),
+                None => {
+                    bare.insert(spec.package);
+                }
+            }
+        }
+
+        let mut conflicts: Vec<String> = versioned
+            .iter()
+            .filter(|(package, _)| bare.contains(*package))
+            .map(|(package, keys)| {
+                format!(
+                    "override `{package}` applies to every version of the package, so it \
+                     conflicts with the versioned override(s) `{}`",
+                    keys.join("`, `")
+                )
+            })
+            .collect();
+        if conflicts.is_empty() {
+            return Ok(());
+        }
+        conflicts.sort_unstable();
+        anyhow::bail!("{} - remove one or the other", conflicts.join("; "));
     }
 
     // no overrides or top-level metadata when workspace is present
@@ -239,6 +286,70 @@ mod tests {
         assert_eq!(
             manifest, loaded_manifest,
             "manifest loaded from file does not match original manifest"
+        );
+    }
+
+    #[test]
+    fn override_keys_may_carry_a_version() {
+        let manifest = Manifest::from_toml(
+            r#"
+[overrides]
+"foo:bar@0.1.0" = { path = "bar-0.1.0" }
+"foo:bar@0.2.0" = { path = "bar-0.2.0" }
+"foo:baz" = { path = "baz" }
+"#,
+        )
+        .expect("versioned override keys should be accepted");
+        assert_eq!(manifest.overrides.unwrap().len(), 3);
+    }
+
+    #[test]
+    fn override_keys_conflict_when_bare_and_versioned() {
+        let err = Manifest::from_toml(
+            r#"
+[overrides]
+"foo:bar" = { path = "bar" }
+"foo:bar@0.1.0" = { path = "bar-0.1.0" }
+"#,
+        )
+        .expect_err("a bare key alongside a versioned one is ambiguous");
+        let err = format!("{err:#}");
+        assert!(err.contains("foo:bar@0.1.0"), "unexpected error: {err}");
+    }
+
+    #[test]
+    fn override_key_conflicts_are_all_reported_in_a_stable_order() {
+        // Two conflicting packages: both must appear, and always in the same order, rather than
+        // whichever the underlying map happened to yield first.
+        let err = Manifest::from_toml(
+            r#"
+[overrides]
+"zzz:two" = { path = "z" }
+"zzz:two@0.2.0" = { path = "z2" }
+"aaa:one" = { path = "a" }
+"aaa:one@0.1.0" = { path = "a1" }
+"#,
+        )
+        .expect_err("both packages conflict");
+        let err = format!("{err:#}");
+        let aaa = err.find("aaa:one").expect("aaa:one should be reported");
+        let zzz = err.find("zzz:two").expect("zzz:two should be reported");
+        assert!(aaa < zzz, "conflicts should be sorted: {err}");
+    }
+
+    #[test]
+    fn override_keys_must_parse() {
+        let err = Manifest::from_toml(
+            r#"
+[overrides]
+"not a package ref" = { path = "bar" }
+"#,
+        )
+        .expect_err("an unparseable override key should be rejected");
+        let err = format!("{err:#}");
+        assert!(
+            err.contains("invalid override key"),
+            "unexpected error: {err}"
         );
     }
 }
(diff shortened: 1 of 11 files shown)


mkatychev:
Hi @Aditya1404Sal, thanks for the submission.
Before I take a look at the code, could you pare down your PR description? The description is a bit too much context and reads as if an LLM wrote it.

Aditya1404Sal:
> Hi @Aditya1404Sal, thanks for the submission. Before I takr a look at the code, could you pare down your PR description? The description is a bit too much context and reads as if an LLM wrote it.

Sorry for that, this PR description is indeed ai-assisted. I've rewritten it to my taste.

mkatychev:
> According to the rules governing wit, there's nothing that actively restricts a scenario in which
a wit world can name more than one version of the same package

@lukewagner apologies for pinging, just wanted to get your feedback and see if this is something we want to affirm for package distribution.

macovedj:
Was chatting a bit with @mkatychev.  I'm not sure this is desired behavior.  First off, in your example, the major version is 0, so the bump from 0.2, to 0.3 should signify a breaking change.  I think at the very least we'd probably want to not resolve multiple versions when the change is breaking.  But even if not breaking, I think a wit/interface change is meant to be more intentional than just bumping a library.  For example, the bump from wasi 0.2 to 0.3 completely changes dependence on wasi:io.

All this said, I do believe that package versions are currently treated as a semver carat, so if a new non breaking version is published, you'll automatically get it, which doesn't feel so far off from the spirit of the PR.  I am curious though if you've tried using a file path dependency.  It seems like that could potentially meet your needs @Aditya1404Sal without the potential baggage of supporting resolution for multiple package versions.

crates/wasm-pkg-core/src/manifest.rs
```diff
@@ -79,10 +80,56 @@ impl Manifest {
...
+    ///
+    /// Runs when a `wkg.toml` is loaded (see [`validate`](Self::validate)), and again when
+    /// resolving, since a `Manifest` built directly in Rust code skips the load step.
+    pub(crate) fn validate_override_keys(&self) -> Result<()> {
+        let Some(overrides) = self.overrides.as_ref() else {
+            return Ok(());
+        };
+        // `overrides` is a map, so walk it in a stable order
+        let mut sorted_keys: Vec<&String> = overrides.keys().collect();
+        sorted_keys.sort_unstable();
```
mkatychev, line 99: can this be a `BTreeSet<&str>` to skip the explicit sort?

mkatychev:
@macovedj, had some discussion with @salmans @vados-cosmonic about it and it feels like this is something we should allow this; more context here: https://github.com/bytecodealliance/meetings//blob/main/SIG-Packaging/2026/2026-08-19.md#multiple-versions-of-the-same-pacakge-as-dependencies-for-one-package

crates/wasm-pkg-core/src/resolver.rs
```diff
@@ -104,6 +104,67 @@ impl FromStr for RegistryPackage {
...
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct DependencyKey {
+    /// The package the dependency refers to.
+    pub package: PackageRef,
+    /// The version this dependency was requested at.
+    ///
+    /// `None` means "every version". This only happens for a bare override key (no `@version`),
+    /// which applies to every version the WIT names.
+    pub version: Option<VersionReq>,
+}
```
mkatychev, line 121: Could we reuse the `PackageSpec` type instead of defining `DependencyKey`?
https://github.com/bytecodealliance/wasm-pkg-tools/blob/bc0ece420f8c3e63bb933aa5d53cce8fbb33b312/crates/wasm-pkg-common/src/package.rs#L83-L86

Aditya1404Sal: Done

crates/wasm-pkg-core/src/manifest.rs
Aditya1404Sal: changed it

crates/wasm-pkg-core/src/manifest.rs
```diff
@@ -79,10 +80,55 @@ impl Manifest {
...
         self.validate_workspace_exclusivity()?;
+        self.validate_override_keys()?;
         // Add new validation rules with `self.validate_*()?;`
         Ok(())
     }
 
+    /// Checks that override keys parse and that no package is covered by both a bare and a
+    /// versioned key.
+    ///
+    /// Runs when a `wkg.toml` is loaded (see [`validate`](Self::validate)), and again when
```
mkatychev: https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html
```suggestion
    /// Runs when a `wkg.toml` is loaded (see [`Self::validate`]), and again when
```

crates/wasm-pkg-core/src/manifest.rs
```diff
@@ -79,10 +80,55 @@ impl Manifest {
...
+    /// resolving, since a `Manifest` built directly in Rust code skips the load step.
```
mkatychev: Actually this section should be removed since it implies assumptions about a potential parent scope
```suggestion
```

mkatychev (commented):
@Aditya1404Sal I have been away for the last two weeks so I appreciate the (involuntary!) patience

Aditya1404Sal:
@mkatychev -- no worries at all :) hope you had some quality time!

crates/wasm-pkg-core/tests/fetch.rs
```diff
@@ -158,6 +158,209 @@ async fn test_transitive_local(#[values(OutputType::Wasm, OutputType::Wit)] outp
...
+                },
+            ),
+            (
+                "my:local@0.2.0".to_string(),
+                Override {
+                    path: Some(fixture_path.join("local-dep-0.2.0/wit")),
+                    version: None,
+                },
+            ),
+        ])),
```
mkatychev: are we able to use `Manifest::from_toml` here to roll the comment and concrete object into one? 
https://github.com/bytecodealliance/wasm-pkg-tools/blob/bc0ece420f8c3e63bb933aa5d53cce8fbb33b312/crates/wasm-pkg-core/src/manifest/workspace.rs#L237-L244

crates/wasm-pkg-core/src/resolver.rs
```diff
@@ -390,35 +408,74 @@ impl<'a> DependencyResolver<'a> {
...
+    }
+
+    /// Recorded before the dependency itself is added, so that versioned requests for the same
+    /// package arriving later are skipped. Only overrides get here: an unversioned WIT import is
+    /// a single dependency with no version, not an override covering every version.
+    fn record_any_version_override(&mut self, key: &PackageSpec) {
+        if key.version.is_none() {
+            self.any_version_overrides.insert(key.package.clone());
+        }
+    }
```
mkatychev: I'm a little confused as to the limits of this step, can a key override other overrides?

If we should probably forbid having multiple overrides matching to the same value.

EDIT: I realise that `override_key_conflicts_are_all_reported_in_a_stable_order` touches on this but we should be exiting early without having to resort to asking things such as `is_already_superseded` and `has_superseded_resolution`

mkatychev: It may be better to collect all overrides and ensuring they do not intersect before applying the overrides.

crates/wasm-pkg-core/src/resolver.rs
```diff
@@ -104,6 +104,22 @@ impl FromStr for RegistryPackage {
     }
 }
 
+/// Returns whether resolving `key` also satisfies `other`. A key without a version (`foo:bar`)
+/// covers every version of its package; `foo:bar@0.1.0` covers only itself.
+fn key_supersedes(key: &PackageSpec, other: &PackageSpec) -> bool {
+    key.package == other.package && (key.version.is_none() || key.version == other.version)
+}
```
mkatychev: we should probably have this and `key_requirement` be methods on `PackageSpec` and expose methods that allow us to test for intersections (similar to https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.intersection) since something like `local-dep@*` is a range whereas `local-dep@=0.1.0` is better thought of as a point.

mkatychev: We should also be unambiguous about distinguishing a `PackageSpec { .., version: Some(..) }` where a package has NO version (should not match any version) versus one where it can match ALL versions since they can mean completely different things.

Perhaps a `PackageRequirement`?
```rs
pub struct PackageReq {
    pub package: PackageRef,
    pub version: VersionReq,
}
```

Though I'm not sure if that is valid either, this definitely feels like a design issue if we're having to reach for things like `take_superseded_dependencies`.

mkatychev: * We should ideally know inside `DependencyResolution` if a package requirement has already been resolved locally or not: https://github.com/bytecodealliance/wasm-pkg-tools/blob/bc0ece420f8c3e63bb933aa5d53cce8fbb33b312/crates/wasm-pkg-core/src/resolver.rs#L172-L179
* different resolution requirements likely need to be applied to a local or registry resolution ( forbidding unversioned packages in registries should still apply) but one should be able to tell whether a `LocalResolution` resolved not just to a path but an explicit version (or explicit lack thereof)



Perhaps prepopulating `DependencyResolutionMap` with a new explicit variant might help: `DependencyResolution::Unresolved`:
https://github.com/bytecodealliance/wasm-pkg-tools/blob/bc0ece420f8c3e63bb933aa5d53cce8fbb33b312/crates/wasm-pkg-core/src/resolver.rs#L664-L665

Ultimately we want to be able to resolve requirements without worrying about order of application, this means being able to identify overlaps in overrides _before_ matching  requirements with resolutions; stated more generically: "we want a dependency _requirement_ to be able to path to only one dependency _resolution_".

mkatychev: It may be a good idea to look into how [`PublishPlan` uses acyclic graphs](https://github.com/bytecodealliance/wasm-pkg-tools/pull/215) since we want only one `requirement -> resolution` per requirement (but a resolution can meet more than one requirement):
https://github.com/bytecodealliance/wasm-pkg-tools/blob/bc0ece420f8c3e63bb933aa5d53cce8fbb33b312/crates/wasm-pkg-core/src/resolver.rs#L865-L871

PR #253, opened 2026-08-18
doc(oci): add examples of calling `wkg oci pull` by SHA

### LINKED ISSUE #248: Consider adding GitHub example to README.md
As someone who doesn't know almost anything about OCI and mostly just wanted to "fetch the `.wasm`" given the information on the GitHub package page that contains the `$ docker pull ghcr.io/...` command string, my first few attempts to use `wkg oci pull` failed with somewhat confusing error messages.  E.g., just swapping `wkg oci pull` for `docker pull` produces the error `Image manifest not found: no entry found in image index manifest matching client's default platform`.  IIUC, what I should have done is swap out the `sha256-....` in the GitHub text with the version string.

Based on this, I was thinking that it might be nice to (1) support the `sha256-...` tag if that makes sense or more-helpful error message if not, (2) include a `ghcr.io` example in `README.md` that takes you from the info that's in the GitHub page to the `wkg` command.  Thanks for the consideration!

mkatychev:
This adds documentation and OCI URI ebnf link to show how an image is pulled using the digest hash.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,14 @@ Publish a package to the configured registry:
 wkg publish path/to/component.wasm
 ```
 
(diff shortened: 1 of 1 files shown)


vados-cosmonic (approved):

vados-cosmonic:
Thanks for updating the docs!

